PR #1, opened 2022-11-14
Use cross-env for environment variables

timtmok:
I tried to build the visual editor on Windows and the script is setting the environment variables in a way that won't work on Windows. Using `cross-env` was a quick way to get it to work vs. editing the vite config to add equivalent arguments.


timtmok:
@jjallaire This is what I did really quickly to get it working locally while I update the IDE build. Feel free to change it up if you wanted to handle Windows differently.

jjallaire:
The idea here is that we should eliminate the `rstudio-` scripts entirely from package.json (and that you should set those environment variables in ` build.xml`). Would that work on Windows?

timtmok:
Yes, I see that `<exec>` can set environment variables. One less dependency!

PR #293, opened 2023-10-03
remove --no-watch-inputs from vscode quarto preview

chendaniely:
The `Quarto: Preview` is not watching input changes in VSCode. This is an attempt to get the VSCode extension to behave closer to the CLI.


jjallaire:
It purposely doesn't watch changes so that there is a distinction between save and render (in the case that render is extremely expensive e.g. a few minutes it's really terrible to not provide the user a way to save and not render). If you want the render on save behavior you can opt back into it globally, per-workspace, or per document as described here: https://quarto.org/docs/tools/vscode.html#render-on-save

PR #431, opened 2024-05-03
Some maintenance on VS Code extension

juliasilge:
This PR does not change any functionality but instead makes some updates to resolve errors and problems that arise when trying to work on the extension.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -28,37 +28,19 @@
   ],
   "private": true,
   "engines": {
-    "vscode": "^1.66.0"
+    "vscode": "^1.75.0"
   },
   "main": "./out/main.js",
   "browser": "./browser.js",
   "activationEvents": [
-    "onLanguage:quarto",
-    "onLanguage:mermaid",
-    "onLanguage:dot",
     "onLanguage:r",
-    "onCommand:quarto.preview",
-    "onCommand:quarto.previewFormat",
-    "onCommand:quarto.renderDocument",
-    "onCommand:quarto.renderProject",
     "onCommand.quarto.editInSourceMode",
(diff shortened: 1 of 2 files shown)


apps/vscode/package.json
```diff
@@ -28,37 +28,19 @@
   ],
   "private": true,
   "engines": {
-    "vscode": "^1.66.0"
+    "vscode": "^1.75.0"
```
juliasilge, line 31: The changes in this file are to resolve these problems:

activation-events

VS Code now does [implicit activation](https://github.com/microsoft/vscode-docs/blob/vnext/release-notes/v1_74.md#implicit-activation-events-for-declared-extension-contributions) when your extension provides a language or command.

apps/vscode/tsconfig.json
```diff
@@ -7,7 +7,6 @@
 			"ES2021"
 		],
 		"sourceMap": true,
-		"rootDir": "src",
```
juliasilge, line 10: This change is to avoid the error:

```
File '/Users/juliasilge/Work/rstudio/quarto/apps/vscode/build.ts' is not under 'rootDir' '/Users/juliasilge/Work/rstudio/quarto/apps/vscode/src'. 'rootDir' is expected to contain all source files.
  The file is in the program because:
    Matched by default include pattern '**/*'
```

I did some research on this problem, and I believe we don't need to specify a `rootDir` here anymore.

juliasilge:
I ran the extension with these changes and there were no problems activating the extension or doing some basic previews.

apps/vscode/tsconfig.json
cscheid: I'm ok removing this so we get a clean problems pane. We can revert it if needed.

PR #434, opened 2024-05-07
Add GH action to generate `.vsix` artifact

juliasilge:
Let's see how many tries it takes me to get this to work. 🤞

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -35,5 +35,5 @@ jobs:
       - name: Upload extension vsix file as a workflow artifact
         uses: actions/upload-artifact@v3
         with:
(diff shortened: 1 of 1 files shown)


juliasilge:
Only three tries?! Maybe a record for me with GH actions.

juliasilge:
You can go to the [page for an action run](https://github.com/quarto-dev/quarto/actions/runs/8992442652) to find the link for the `.vsix` and download it. We could eventually publish these as part of a release to surface them more, but for our development work, I think this is the best option.

After that, you can install it via "Extensions: Install from VSIX..." command.

cscheid:
I agree that not advertising releases is still the best way to go about this.

cscheid (approved):
Looks good!

PR #435, opened 2024-05-08
Render output in a temp directory, when in an R package

juliasilge:
This PR updates the preview behavior so by default output files go to a temporary directory when the workspace is an R package.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -1073,6 +1073,13 @@
           "default": true,
           "markdownDescription": "Reveal the preview panel after document render."
         },
+        "quarto.render.rPackageOutputDirectory": {
+          "order": 15,
+          "scope": "window",
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Render output files in a temporary directory, when in an R package."
+        },
         "quarto.visualEditor.fontSize": {
           "order": 31,
           "scope": "resource",
--- a/apps/vscode/src/providers/preview/preview-util.ts
+++ b/apps/vscode/src/providers/preview/preview-util.ts
@@ -59,8 +59,11 @@ export function isQuartoShinyKnitrDoc(
   
 }
 
-export async function isRPackage() {
+export async function isRPackage(): Promise<boolean> {
   const descriptionLines = await parseRPackageDescription();
+  if (!descriptionLines) {
+    return false;
+  }
   const packageLines = descriptionLines.filter(line => line.startsWith('Package:'));
   const typeLines = descriptionLines.filter(line => line.startsWith('Type:'));
   const typeIsPackage = (typeLines.length > 0
--- a/apps/vscode/src/providers/preview/preview-util.ts
+++ b/apps/vscode/src/providers/preview/preview-util.ts
@@ -59,6 +59,30 @@ export function isQuartoShinyKnitrDoc(
   
 }
 
+export async function isRPackage() {
+  const descriptionLines = await parseRPackageDescription();
+	const packageLines = descriptionLines.filter(line => line.startsWith('Package:'));
+	const typeLines = descriptionLines.filter(line => line.startsWith('Type:'));
+	const typeIsPackage = (typeLines.length > 0
+		? typeLines[0].toLowerCase().includes('package')
+		: false);
+	const typeIsPackageOrMissing = typeLines.length === 0 || typeIsPackage;
+	return packageLines.length > 0 && typeIsPackageOrMissing;
+}
+
+async function parseRPackageDescription(): Promise<string[]> {
+	if (vscode.workspace.workspaceFolders !== undefined) {
+		const folderUri = vscode.workspace.workspaceFolders[0].uri;
+		const fileUri = vscode.Uri.joinPath(folderUri, 'DESCRIPTION');
+		try {
+			const bytes = await vscode.workspace.fs.readFile(fileUri);
+			const descriptionText = Buffer.from(bytes).toString('utf8');
+			const descriptionLines = descriptionText.split(/(\r?\n)/);
+			return descriptionLines;
+		} catch { }
+	}
+	return [''];
+}
 
 export async function renderOnSave(engine: MarkdownEngine, document: TextDocument) {
   // if its a notebook and we don't have a save hook for notebooks then don't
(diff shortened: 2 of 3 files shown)


apps/vscode/package.json
```diff
@@ -1073,6 +1073,13 @@
           "default": true,
           "markdownDescription": "Reveal the preview panel after document render."
         },
+        "quarto.render.rPackageOutputDirectory": {
+          "order": 15,
+          "scope": "window",
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Render output files in a temporary directory, when in an R package."
+        },
```
juliasilge, line 1082: This is new configuration, so users can turn off this behavior if they prefer:

Screenshot 2024-05-08 at 4 08 13 PM


I want to argue that this should be `true` by default, because otherwise, R package developers need to add all these output files to both `.gitignore` and `.Rbuildignore`.

apps/vscode/src/providers/preview/preview-util.ts
```diff
@@ -59,6 +59,30 @@ export function isQuartoShinyKnitrDoc(
   
 }
 
+export async function isRPackage() {
```
juliasilge, line 62: This code is similar to what we are using elsewhere to detect if we are in an R package; it is working well for us so far. Consider it analogous `isQuartoShinyDoc()` in this same file, although it applies to the _workspace_ rather than a file. We believe it is reasonable to make assumptions about folks using workspaces for R packages.

juliasilge (commented):
In my testing, this works well for files at the top level of an R package like README and NEWS. We'll need to follow up for vignettes, because although the final file is written to the temporary directory, these kinds of file are still getting written to the directory where the vignette lives:

Screenshot 2024-05-08 at 4 19 21 PM

apps/vscode/package.json
cscheid: I would recommend doing a bit of work ahead of time to make sure that this is not going to create other bugs. In an R package, will users mostly preview `README.md` files and vignettes? In that case, I wonder if the right behavior is to use `--output-dir` only in these special cases. 

Like we discussed, relocating files for an .html target is unlikely to be robust because of things like CSS resource paths; in that case, we tend to recommend to use `--output-dir` together with [`embed-resources`](https://quarto.org/docs/output-formats/html-basics.html#self-contained)

cscheid: With that said, this is a pain in the neck to test in an automated fashion. It might be good enough to test this by making a mock test harness over at https://github.com/quarto-dev/quarto-cli. I can help with that; I think it's prudent to do it. Because `quarto-cli` won't generally know the ways in which the extension calls `quarto render`, `quarto preview` etc, we have a pretty big regression risk unless we do something about it.

apps/vscode/src/providers/preview/preview-util.ts
cscheid: The code looks good, but it doesn't look like auto-formatting is working in this repo, and this uses tabs 😱.

apps/vscode/src/providers/preview/preview-util.ts
```diff
@@ -59,6 +59,30 @@ export function isQuartoShinyKnitrDoc(
...
+async function parseRPackageDescription(): Promise<string[]> {
+	if (vscode.workspace.workspaceFolders !== undefined) {
+		const folderUri = vscode.workspace.workspaceFolders[0].uri;
+		const fileUri = vscode.Uri.joinPath(folderUri, 'DESCRIPTION');
+		try {
+			const bytes = await vscode.workspace.fs.readFile(fileUri);
+			const descriptionText = Buffer.from(bytes).toString('utf8');
+			const descriptionLines = descriptionText.split(/(\r?\n)/);
+			return descriptionLines;
+		} catch { }
```
cscheid, line 82: Are we sure we want to eat all exceptions here?

cscheid (commented):
There's a `catch { }` clause that I think we should be more careful about before merging, but this otherwise looks good.

apps/vscode/src/providers/preview/preview-util.ts
juliasilge: I'm very happy to help get some minimal `settings.json` set up in this repo, or of course for you to do it and then adjust. If I and/or other people are going to be contributing more, it probably makes sense to do that pretty soon.

apps/vscode/src/providers/preview/preview-util.ts
juliasilge: I believe so, yes! First, we try to parse the `DESCRIPTION` file. If it works, we return the contents in `descriptionLines` and if anything goes wrong, we return `''` from this function. I'll add an early return with `if(!descriptionLines)` to `isRPackage()` to improve it a bit.

apps/vscode/package.json
juliasilge: > In that case, I wonder if the right behavior is to use `--output-dir` only in these special cases.

I'm going to get some more feedback on this from folks, but my current belief is that there are basically _no cases_ where you want to check in generated code such as HTML, libs, etc, for an R package.

The one exception is `README.Rmd` being rendered to `README.md` but a) none of that works for people right now because of `output: github_document` and b) we can encourage them to use `devtools::build_readme()` for this purpose. If we want to be more forward thinking, we could _exclude_ `README.Rmd` and `README.qmd` from this treatment, rather than make a list of files to _include_ for this treatment.

apps/vscode/src/providers/preview/preview-util.ts
cscheid: I think it makes sense to use whatever `settings.json` is common across the rest of the company. `quarto-cli` is a Deno project so it uses `deno.jsonc` instead. I'd be happy for `quarto` to take that `settings.json` in a separate PR.

apps/vscode/src/providers/preview/preview-util.ts
cscheid: Ok! I've been bitten enough by stuff like this that I try to always catch specific exceptions, but we don't have any explicit recommendations (nor are we running eslint here, I don't think), so your explanation is great, thanks!

apps/vscode/package.json
cscheid: > I'm going to get some more feedback on this from folks, but my current belief is that there are basically no cases where you want to check in generated code such as HTML, libs, etc, for an R package.

I believe that! Really my only concern is that `quarto preview` itself might break under common use cases in R packages, and I'd hate for us to enable something by default that doesn't work well in these scenarios.

cscheid:
This looks good from my end, so feel free to squash+merge it or let me know if you have no further changes to make!

PR #436, opened 2024-05-10
Add `executeSelection` method to the extension host cell executor

juliasilge:
This is just draft for now! I need to get some feedback for this.

### DIFF
--- a/apps/vscode/src/host/index.ts
+++ b/apps/vscode/src/host/index.ts
@@ -34,6 +34,19 @@ export interface HostWebviewPanel extends vscode.Disposable {
   readonly onDidDispose: vscode.Event<void>;
 }
 
+export interface HostStatementRangeProvider {
+  provideStatementRange(
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    token: vscode.CancellationToken
+  ): vscode.ProviderResult<HostStatementRange>;
+}
+
+export interface HostStatementRange {
+  readonly range: vscode.Range;
+  readonly code?: string;
+}
+
 export interface ExtensionHost {
 
   // code execution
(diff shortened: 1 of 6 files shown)


jjallaire:
From my standpoint this look like it does the quarto extension bits exactly right.

apps/vscode/src/host/index.ts
```diff
@@ -73,14 +91,20 @@ function defaultExtensionHost() : ExtensionHost {
...
-       return languages.filter(language => knitr || !visualMode || (language !== "python"));
+      // jupyter python (as distinct from knitr python) doesn't work in visual mode b/c
+      // jupyter.execSelectionInteractive  wants a text editor to be active
+      return languages.filter(language => knitr || !visualMode || (language !== "python"));
     },
     cellExecutorForLanguage,
+    // in the default extension host, this is a noop:
+    registerStatementRangeProvider: (
+      engine: MarkdownEngine,
+		): vscode.Disposable => {
```
DavisVaughan, line 105: Formatting seems a little strange here

DavisVaughan (approved):
LGTM in relation to how this hooks up to the other side

apps/vscode/src/host/index.ts
juliasilge: It would be good to get some automatic formatting on this repo @cscheid, like maybe add this?

```json
"[typescript]": {
		"editor.defaultFormatter": "vscode.typescript-language-features",
		"editor.formatOnSave": true
	}
```

cscheid: Yes. It'll cause a bit of source control churn to do it this late in the repo, but better late than never.

cscheid: (It's fixed on `main` now.)

cscheid:
We'll deal with the autoformatting churn in future PRs. This looks ready to go 🚀

Issue #430, opened 2024-05-01
AGPL prohibits VS Code extension usage at many companies


bruceskarin:
The selection of [AGPL for the VS Code extension](https://github.com/quarto-dev/quarto/blob/main/apps/vscode/LICENSE) is not consistent with the rest of the Quarto ecosystem, which is now released under a [MIT license](https://github.com/quarto-dev/quarto-cli/blob/main/COPYRIGHT). Switching to a more permissive license would greatly improve adoption of Quarto across enterprises.

bruceskarin:
Hi @jjallaire, I'm huge fan of RStudio and Quarto. We use both at my work, but they treat AGPL like the third rail and won't make any exceptions. We have to ask for one for LGPL, but both GPL and MIT don't require any. I noticed how Quarto was already chanced from from GPL to MIT, so I'm hoping the same can be done here for the extension. Obviously feel free to revise or replace with whatever you think best serves Quarto. Many thanks to you and the team that make all this amazing software! -Bruce

PR #478, opened 2024-07-02
License updated to match Quarto ecosystem

bruceskarin:
Closes Issue https://github.com/quarto-dev/quarto/issues/430
The selection of AGPL for the VS Code extension is not consistent with the rest of the Quarto ecosystem, which is now released under a MIT license. Switching to same license would greatly improve adoption of Quarto across enterprises.


cscheid:
Appreciate the PR, but we can't make this change because there's code in RStudio that cannot be licensed as MIT and some of that code is in this codebase. It's a bigger change that involves more substantial coordinate across the company, and not one we're ready to make.

PR #503, opened 2024-08-02
Look for Quarto in the app root (for Positron)

jmcphers:
This change allows the Quarto extension to discover a version of Quarto installed (bundled with) Positron. See https://github.com/posit-dev/positron/pull/4212 for details on how Quarto is bundled with Positron.

Since Positron installation locations can vary, and we don't want the installed version of Quarto to take precedence over other installations, the approach taken here is to add the `quarto/bin` folder of the app root to the list of paths to scan for Quarto.

### DIFF
--- a/apps/vscode/src/main.ts
+++ b/apps/vscode/src/main.ts
@@ -55,6 +55,8 @@ export async function activate(context: vscode.ExtensionContext) {
   const quartoContext = initQuartoContext(
     quartoPath,
     workspaceFolder,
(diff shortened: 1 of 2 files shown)


petetronic:
Confirmed that when this change is installed over the quarto extension in positron, it does locate the bundled version of quarto cli in a version of Positron built from PR 4212.

coatless:
Not sure who to tag for a review, but this would be nice to get in! @cderv @cscheid

aisi-inspect (approved):

PR #510, opened 2024-08-06
Use a task queue to ensure code blocks are executed sequentially

### LINKED ISSUE #509: `execute()` hook for Positron can run blocks out of order
See https://github.com/posit-dev/positron/issues/4231

Check out the following video. If you "run all above" and then _very quickly_ run another code block, then the blocks have the potential to be run out of order.

https://github.com/user-attachments/assets/8eb2083c-5d33-4fa4-b96a-baa2e6da8ca3

This is due to the way this `execute()` hook works for Positron

https://github.com/quarto-dev/quarto/blob/06678d55143a7d6de6c7f0232db0c5dcf8392ca6/apps/vscode/src/host/hooks.ts#L58-L67

Note that the `await` happens _within the loop_ at each iteration. That means that after the first `await`, we can switch away to some other code - in particular someone else could call `execute()` and we could execute _their_ code before we finish executing all of the blocks that we need to run. 

@lionel- and I have a patch that only performs the `await` on _the last_ code block. That ensures that we can sequentially fire off all of the `block`s without the potential of having someone else execute code in between two of our blocks. I'll send in a PR for this.

``````

---
title: "Untitled"
format: html
---


```{r}
library(tidyverse)
library(phyloseq)
library(decontam)
```


```{r}
x <- 1
```

```{r}
y <- x
```

```{r}
z <- y
```

```{r}
a <- z
```


```{r}

```


``````

### PR DESCRIPTION
**Pull request by DavisVaughan**
Closes #509 

Joint work with @lionel-

This ensures that all of the blocks are sent over in a single sequential batch, with no way for anyone else to execute their own block between two of ours, which would cause ordering issues

### CODE REVIEW

**Comment by DavisVaughan** [comment]
While I have not tested it, I believe this problem also exists for the Jupyter executor, as it also `await`s in a loop, giving someone else the opportunity to inject a block between two executions

https://github.com/quarto-dev/quarto/blob/06678d55143a7d6de6c7f0232db0c5dcf8392ca6/apps/vscode/src/host/executors.ts#L63-L78

**Comment by lionel-** [comment]
Since this is tricky to get right, it seems like vsc-quarto should be in charge of executing the commands so that it can make sure these are run "atomically", i.e. all commands are run before yielding back to the event loop. To this end there could be a new API where language extensions provide an `executeCode()` callback to execute a block that vsc-quarto would then call in a loop structured similarly to the one we implemented here.

File: apps/vscode/src/host/hooks.ts [hunk 1]
```diff
@@ -56,14 +56,32 @@ export function hooksExtensionHost() : ExtensionHost {
...
+              // as opposed to `await`ing at each iteration within the loop. This prevents
+              // someone else from executing their own block while we `await`, which would
+              // lead to ordering issues (https://github.com/posit-dev/positron/issues/4231).
+              for (const block of blocks) {
+                executeCode(block);
+              }
+
+              if (lastBlock !== undefined) {
+                await executeCode(lastBlock);
+              }
```
cscheid, line 84: This looks a little scary to me; do I read it right that we are trusting that executeCode will naturally maintain the correct execution order? Isn't this asking for a race condition to go bad? Should there be a `executeCodeArray` method that offers the right guarantee, and internally does this work?

I'm happy to take our standard "you break it, you buy it" contract here, but it does strike me as a potentially dangerous change.

cscheid:
(just letting you know that this change looks a lot safer now, from what I can tell!)

apps/vscode/src/host/manager.ts
```diff
@@ -0,0 +1,77 @@
+/*
+ * manager.ts
```
lionel-, line 2: `execute-queue.ts` would be more telling?

apps/vscode/src/host/manager.ts
```diff
@@ -0,0 +1,77 @@
...
+
+import { Disposable } from 'vscode';
+import { TaskCallback, TaskQueue } from './queue';
+
+/**
+ * Manager of multiple task queues
+ * 
+ * A singleton class that constructs and manages multiple `TaskQueue`s, identified by their `key`.
+ */
+export class TaskQueueManager implements Disposable {
```
lionel-, line 24:
```suggestion
export class ExecuteQueue implements Disposable {
```

?

lionel- (commented):
I like this new setup!

I think we could use an existing async queue? In the `positron-r` extension we use [PQueue](https://github.com/posit-dev/positron/blob/b1b377b0e3b3b03ff3f80980acd508526c307ec0/extensions/positron-r/src/session.ts#L46).

DavisVaughan:
@lionel- it seems like PQueue was actually exactly what we needed. I took a look at their implementation and it seems _extremely_ similar to my `queue.ts` approach, I basically reimplemented it without knowing. Can you take one more look please?

PR #537, opened 2024-09-11
suppress background highlight on first line

kevinushey:
Some Ace highlight rules will also give tokens a background highlight color. For example, Stan's highlight rules (in RStudio) give `var` a red background highlight. Unfortunately, these highlight rules also apply to tokens on the first line, and so we can end up with a chunk like so:

Screenshot 2024-09-11 at 3 09 37 PM

This PR alleviates that by forcing `background: none !important` in the CSS for panmirror.


cscheid:
Is there any reason we'd ever want the first line to be highlighted?

jjallaire (approved):

kevinushey:
> Is there any reason we'd ever want the first line to be highlighted?

I don't think so, but it's plausible that we'll want to some day adjust the treatment of the chunk header in the future? That seems like an overall larger project though.

PR #542, opened 2024-09-13
Publish extension via GH action on GH release

juliasilge:
This PR sets up infrastructure to publish the extension to the two marketplaces (OpenVSX and the Visual Studio Marketplace) via GH action via doing a GH _release_. I'll make a few notes on the PR for us to use for discussion moving forward.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,6 +2,9 @@
 name: Build VS Code Extension
 
 on:
+  release:
+    types:
+      - published
   push:
     branches:
       - main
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yaml
```diff
@@ -2,6 +2,9 @@
 name: Build VS Code Extension
 
 on:
+  release:
+    types:
+      - published
```
juliasilge, line 7: I set this up to use a _release_ but alternatively we could [use a tag in the way the Shiny team does](https://github.com/posit-dev/shiny-vscode/blob/main/.github/workflows/publish.yml).

.github/workflows/ci.yaml
```diff
@@ -37,3 +40,34 @@ jobs:
...
+          node-version: latest
+      - run: yarn install --immutable --immutable-cache --check-cache
+      - uses: actions/download-artifact@v4
+        with:
+          name: quarto-vscode-${{ github.sha }}
+
+      - name: Publish to Open VSX Registry
+        uses: HaaLeo/publish-vscode-extension@v1
+        with:
+          pat: ${{ secrets.OPEN_VSX_TOKEN }}
```
juliasilge, line 65: This is the first of two tokens we will need added to the repo secrets to make this work.

.github/workflows/ci.yaml
```diff
@@ -37,3 +40,34 @@ jobs:
...
+          extensionFile: ${{ needs.package_extension.outputs.vsixPath }}
+
+      - name: Publish to Visual Studio Marketplace
+        uses: HaaLeo/publish-vscode-extension@v1
+        with:
+          pat: ${{ secrets.VS_MARKETPLACE_TOKEN }}
```
juliasilge, line 71: This is the second of two tokens we will need added to the repo secrets to make this work.

juliasilge (commented):
This action builds the extension one time and then publishes exactly the same file to both OpenVSX and the Microsoft marketplace. There are alternative approaches if we have a reason for another approach.

.github/workflows/ci.yaml
juliasilge: JJ added this a couple of days ago to this repo's secrets. ✅

.github/workflows/ci.yaml
juliasilge: I just added this to this repo's secrets, after being added as a publisher for the Quarto extension. I made a token that expires after 90 days FYI so we will need to replace it at that time. If we feel comfortable, next time I can make one that never expires.

cscheid:
This all looks good to me!

PR #549, opened 2024-09-20
Separate and update GH actions for VS Code extension

juliasilge:
After our successful 🎉 experience today publishing the VS Code extension via GH action, we have a couple of improvements to make.

This PR separate the GH action into two separate ones:

- One that builds the `.vsix` as a way to easily get a dev version into people's hands. This GH action now will run on a push or PR to main at the path `apps/vscode` (not any push/PR to main).
- One that publishes the extension to both the Microsoft and OpenVSX marketplaces. This GH action runs on workflow dispatch (i.e. pushing the button) and will also create a GH release similar to the one [we just made](https://github.com/quarto-dev/quarto/releases). This release will also include the `.vsix`. I did not set up any auto-generation of release notes; for now, we can edit the release notes after it is published. A future improvement here would be to autogenerate these. This action should be run when the extension is totally ready to go, i.e. the changelog is up to date and the version in `package.json` has been bumped.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -2,15 +2,16 @@
 name: Build VS Code Extension
 
 on:
-  release:
-    types:
-      - published
   push:
     branches:
       - main
+    paths:
+      - apps/vscode 
   pull_request:
     branches:
       - main
+    paths:
+      - apps/vscode 
   workflow_dispatch:
 
 jobs:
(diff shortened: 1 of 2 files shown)


juliasilge:
The next person to use the GH action to publish should be aware that it hasn't actually been run yet to do so, especially with respect to the release bit. A little more polish might be needed.

cscheid:
This looks good to me! We can pair on zoom this week to try it out again, or we can wait until your next release to try it. Your call.

juliasilge:
Let's merge this now and then wait until the next release to test it out and see if any further slight changes need to be made. Thanks!

PR #551, opened 2024-09-23
Raw html regression

gtritchie:
Work in progress on fixing https://github.com/rstudio/rstudio/issues/15189. This problem also happens when using the Visual Studio Code extension.

Caused by a change in pandoc behavior introduced in 3.2: https://github.com/jgm/pandoc/issues/9677


gtritchie:
> I think a ton of user documents are going to break if you do -raw-html.

PR #553, opened 2024-09-25
use custom pandoc writer to preserve raw html blocks

### LINKED ISSUE #552: Visual Editor loses raw HTML code blocks
NOTE: This issue is due to a pandoc change, and also impacts the RStudio IDE (https://github.com/rstudio/rstudio/issues/15189). I am working on a fix for both the IDE and the VSCode extension.

### Steps to reproduce the problem

- Using extension version 1.115.0 in VSCode 1.93.1
- Create a new Quarto document, edit in non-visual mode, add following:

````qmd
---
title: "Raw HTML Block Issue"
format: html
---

Some text.

```{=html}
<p>Put <b>HTML</b> <i>here</i>.</p>
```

This is <b>inline HTML</b>.
````

Save the file, then Edit in Visual Mode. It looks ok:

vscode-visual

Now switch back to source (non-visual) mode.

### Describe the problem in detail

The code block is gone, all that is left is the content, but not the surrounding backticks, etc.

````qmd
---
title: "Raw HTML Block Issue"
format: html
---

Some text.

<p>Put <b>HTML</b> <i>here</i>.</p>

This is <b>inline HTML</b>.
````

### Describe the behavior you expected

Raw HTML blocks should not be removed by the Visual Editor.

### PR DESCRIPTION
**Pull request by gtritchie**
Fixes #552

Uses a custom pandoc writer to preserve raw HTML blocks (```{=html}```) when using Visual Editor.

This was due to a Pandoc 3.2+ change that was stripping these blocks when we were converting from the AST to markdown.

Note/question: I updated the CHANGELOG and extension version as part of this; not sure if that's appropriate or should be done separately.

The same change will be made separately in RStudio; it will consume this same `md-writer.lua` script (rstudio pulls in this repo to obtain panmirror so can grab it from there) and `rsession` will make the same substitution that is happening here.

### DIFF
--- a/packages/editor-server/src/server/pandoc.ts
+++ b/packages/editor-server/src/server/pandoc.ts
@@ -55,6 +55,13 @@ import { EditorServerOptions } from './server';
 
 export function pandocServer(options: EditorServerOptions) : PandocServer {
 
+  // work around change in pandoc 3.2+: https://github.com/jgm/pandoc/issues/9677 by
+  // using a custom pandoc writer
+  function substituteCustomMarkdownWriter(format: string): string {
+    const customWriterScript = path.join(options.pandoc.resourcesDir, 'md-writer.lua')
+    return format.replace(/^markdown(?=[-+]|$)/, customWriterScript);
+  }
+
   async function runPandoc(args: readonly string[] | null, stdin?: string) : Promise<string> {
     return pandoc(options.pandoc, args, stdin);
   }
@@ -100,9 +107,9 @@ export function pandocServer(options: EditorServerOptions) : PandocServer {
       return ast;
     },
     async astToMarkdown(ast: PandocAst, format: string, options: string[]): Promise<string> {
-      const markdown = await runPandoc(
+     const markdown = await runPandoc(
         ["--from", "json",
-         "--to", format, ...options],
+         "--to", substituteCustomMarkdownWriter(format), ...options],
          JSON.stringify(ast)
       );
       return markdown;
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 1.116.0 (Not yet released)
+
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: packages/editor-server/src/server/pandoc.ts [hunk 1]
```diff
@@ -55,6 +55,13 @@ import { EditorServerOptions } from './server';
 
 export function pandocServer(options: EditorServerOptions) : PandocServer {
 
+  // work around change in pandoc 3.2+: https://github.com/jgm/pandoc/issues/9677 by
+  // using a custom pandoc writer
+  function substituteCustomMarkdownWriter(format: string): string {
+    const customWriterScript = path.join(options.pandoc.resourcesDir, 'md-writer.lua')
+    return format.replace(/^markdown(?=[-+]|$)/, customWriterScript);
```
cscheid, line 62: This will eat all extensions that might be passed as `format`, right? Is that a problem? I'm not sure about the context around `astToMarkdown`, but it's probably worth checking.

gtritchie: No, I preserve everything after `markdown`, so if the format string comes in as:

`markdown+foo-bar` that gets replaced by `/path/to/md-writer.lua+foo-bar` and so on.

gtritchie: Hmm, let me double check that bold statement.

gtritchie: Double-checked.

regex-test

cscheid: It's been 0 days since a regexp misreading in this factory

gtritchie: Docs on passing extensions to custom writers: https://github.com/jgm/pandoc/blob/main/doc/custom-writers.md#format-extensions

juliasilge (approved):
I am not currently able to comment on the visual editor change, but since this does go out with the VS Code extension, we do want to update the changelog for it and this looks good to me. Thanks!

jjallaire (approved):

Issue #558, opened 2024-10-02
Visual Editor loses raw latex code blocks


gtritchie:
NOTE: This issue is due to a pandoc change, and also impacts the RStudio IDE (https://github.com/rstudio/rstudio/issues/15253). I am working on a fix for both the IDE and the VSCode extension.

See also #552, already fixed.

### Steps to reproduce the problem

- Using extension version 1.115.0 in VSCode 1.93.1
- Create a new Quarto document, edit in non-visual mode, add following:

````qmd
---
title: "Raw tex Block Issue"
---

```{=tex}
\LaTeX
```

```{=latex}
\LaTeX
```

````

Save the file, then Edit in Visual Mode. The latex block looks ok, but the tex block has lost its specialness.

![Image](https://github.com/user-attachments/assets/20decbea-69a5-4730-9779-973d2a051431)

Now switch back to source (non-visual) mode.

### Describe the problem in detail

In Visual Mode, the `tex` block is removed leaving just the contents. The `latex` block looks ok in Visual Mode initially, but the underlying document has changed to:

````qmd
---
title: "Raw tex Block Issue"
---

\LaTeX

\LaTeX
````

### Describe the behavior you expected

Preserve the `latex` raw block as we did before the pandoc change.

Note that `{=tex}` being removed was already present before the pandoc change, so not covered by this issue.

PR #559, opened 2024-10-02
Fix roundtripping of raw latex blocks

gtritchie:
Fixes https://github.com/quarto-dev/quarto/issues/558

Update custom pandoc writer to preserve raw latex blocks ({=latex}) when using Visual Editor.

This was due to a Pandoc 3.2+ change that stripped these blocks when we converted from the AST to markdown.

NOTE, we still don't preserve `{=tex}` fenced blocks, but that was true before the pandoc change. I tried updating the script to also preserve `{=tex}` but it didn't work, so the root cause is elsewhere.


cscheid:
Thanks. This looks good.

With that said, I'm starting to wish we had a test suite here.

cscheid (approved):
lgtm.

jjallaire (approved):

PR #565, opened 2024-10-07
recognize _brand.ya?ml as part of a Quarto project

### LINKED ISSUE #483: yaml intelligence: `_brand.yml`
With the work going on https://github.com/quarto-dev/quarto-cli/tree/feature/brand-yaml, we should add validation and autocompletion to `_brand.yml` just like we have in `_quarto.yml`.

cscheid:
This closes #483.

It requires https://github.com/quarto-dev/quarto-cli/pull/11006 to be merged for it to work.

There's also an issue we haven't had to deal with before, which is how should different versions of the extension interact with different versions of `quarto-cli`. Right now, if the extension is "too new" for the cli, `_brand.yml` will be validated as if it were `_quarto.yml` or `_metadata.yml`. Not horrible, but not ideal. I wonder if we have a mechanism for addressing this.

### DIFF
--- a/packages/quarto-core/src/document.ts
+++ b/packages/quarto-core/src/document.ts
@@ -1,7 +1,7 @@
 /*
  * document.ts
  *
- * Copyright (C) 2023 by Posit Software, PBC
+ * Copyright (C) 2023-2024 by Posit Software, PBC
  * Copyright (c) Microsoft Corporation. All rights reserved.
  *
  * Unless you have received this program directly from Posit Software pursuant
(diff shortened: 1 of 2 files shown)


cscheid:
@juliasilge Nothing too special here, but I wanted to flag that versioning issue I brought up above as something we eventually need to sort out company-wide.

juliasilge:
I may be misunderstanding the particulars of the versioning, but is it possible to check the cli version and gate the new extension behavior on new-enough cli? Like this:

https://github.com/quarto-dev/quarto/blob/20fb64c21c04fd4ca04bae408e1b3aeb53c65dc9/apps/vscode/src/providers/preview/preview.ts#L487

cscheid:
I was thinking more generally about what to do when Posit's product releases don't quite align; is there something we can do as a company to avoid this kind of code? I imagine it'll make everyone's codebases that much more annoying to test.

Nevertheless, yes: right now I should use precisely `QuartoContext.version`. Thanks!

juliasilge:
> is there something we can do as a company to avoid this kind of code?

For tools inside of one ecosystem (R packages, Python packages, etc) there are of course native tools to manage these kinds of version dependencies. It's an interesting question for cross-ecosystem tooling!

packages/quarto-core/src/document.ts
```diff
@@ -100,6 +100,7 @@ export function isQuartoYaml(doc: Document) {
   return (
     doc.languageId === kYamlLanguageId &&
     (doc.uri.match(/_quarto(-.*?)?\.ya?ml$/) ||
+      doc.uri.match(/_brand\.ya?ml$/) ||
```
cscheid, line 103: In case anyone is curious about the lack of a Quarto version check here: if the pattern in `LanguageClientOptions` doesn't match, VS Code never actually handles `_brand.yml` to Quarto's LSP. A version check would be redundant.

PR #567, opened 2024-10-09
Fix how we generate the release name and tag name

juliasilge:
This GH action was not quite configured correctly, so we were evaluating to `''` or something like that for the tag name and GH release name as @cscheid saw during the last release.


cscheid:
LGTM, thanks!

PR #576, opened 2024-10-17
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,0 +1,6 @@
+[codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = .git*,*.svg,i18n,*.lock,*.css,*.min.*,.codespellrc,packages,completion,assets,package.json
+check-hidden = true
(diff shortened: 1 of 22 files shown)


cscheid:
Thanks for the PR and the typo fixes.

I'm reluctant to take in a dependency of a tool no one in our team uses (if we did, it would be supported here already), especially when there's CI consequences that will affect our workflow and we have no experience with them.

yarikoptic:
Hi @cscheid 

the decision is yours to make -- I can only humbly contribute ;)  FWIW - disabling github workflow is easy -- just remove it ;)  maintenance should be minimal if any.

Meanwhile -- rebased, and redid fixing to address conflicts.  Please advise -- I could drop tooling and just leave config (so later runs would not suffer false positives etc) + fixes.

PR #577, opened 2024-10-18
Automatically open QMD files in visual/source editor if specified

isabelizimm:
as part of https://github.com/posit-dev/positron/issues/2933

Should have the following behavior:

- if `editor: XXXX` is specified in yaml at top of file, use that
- if no editor in file yaml, but `editor: XXXX` is specified in `_quarto.yml`/`_quarto.yaml`, use that
- otherwise, use configurable default editor


apps/vscode/src/providers/editor/toggle.ts
```diff
@@ -12,16 +12,99 @@
...
+  const config = vscode.workspace.getConfiguration('quarto').get<string>('defaultEditor');
+  const viewType = config === 'visual' ? VisualEditorProvider.viewType : 'textEditor';
+  vscode.workspace.getConfiguration('workbench').update('editor.defaultView', viewType, true);
+  await vscode.commands.executeCommand("workbench.action.setDefaultEditor",
+    vscode.Uri.file('filename.qmd'),
+    viewType
+  );
+}
+
+export function workspaceHasQuartoYaml() {
```
DavisVaughan, line 65: Maybe more like `modeFromQuartoYaml()`?

DavisVaughan: Also specify return type? 🤔

DavisVaughan: I'd be surprised if there wasn't already something like `getQuartoYaml()` somewhere else

apps/vscode/src/providers/editor/toggle.ts
```diff
@@ -12,16 +12,99 @@
...
+      if (parsedYaml.editor === 'visual' || parsedYaml.editor === 'source') {
+        return parsedYaml.editor;
+      }
+    }
+  }
+
+  return undefined;
+}
+
+export function hasEditorMode(doc: string, mode: string) {
```
DavisVaughan, line 93: I'd also be surprised if there wasn't something like `parseYamlFrontmatter()` somewhere

DavisVaughan: Somewhere around here https://github.com/quarto-dev/quarto/blob/97cc659a3481aca86b04e6c3fb5015d204d83ee8/packages/editor/src/editor/editor.ts#L816

apps/vscode/src/providers/editor/toggle.ts
DavisVaughan: Maybe this?
https://github.com/quarto-dev/quarto/blob/97cc659a3481aca86b04e6c3fb5015d204d83ee8/packages/quarto-core/src/metadata.ts#L44

Combined with `yamlFromMetadata()`?
https://github.com/quarto-dev/quarto/blob/97cc659a3481aca86b04e6c3fb5015d204d83ee8/apps/vscode/src/providers/editor/prefs.ts#L213-L220

DavisVaughan (commented):
Some drive by thoughts

apps/vscode/src/main.ts
```diff
@@ -125,9 +128,46 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    const editorMode = await determineMode(document.getText());
+    if (editorMode && editorMode != config) {
+      suppressOpenHandler = true;
+      const editorOpener = editorMode === 'visual' ? VisualEditorProvider.viewType : 'textEditor';
+      await vscode.commands.executeCommand('workbench.action.closeActiveEditor');
+      await vscode.commands.executeCommand("vscode.openWith",
+        document.uri,
+        editorOpener
+      );
+      return;
```
DavisVaughan, line 160: There seems to be some kind of timing issue here - Even though we are in `onDidOpenTextDocument()`, it is often not the _active_ editor yet, so often it won't actually close the editor, and then `openWith` is a no-op because it thinks that URI is already open.

apps/vscode/src/main.ts
```diff
@@ -125,9 +126,44 @@ export async function activate(context: vscode.ExtensionContext) {
...
   activateCommon(context, host, engine, commands);
+
+  // if positron
+  setEditorOpener();
+
+  vscode.workspace.onDidChangeConfiguration(async (event) => {
+    if (event.affectsConfiguration('quarto.defaultEditor')) {
+      setEditorOpener();
+    }
+  });
```
DavisVaughan, line 137: The disposable returned here needs to be captured so it can be disposed of at some point (like `documentOpenHandler`)

apps/vscode/src/main.ts
```diff
@@ -125,9 +126,44 @@ export async function activate(context: vscode.ExtensionContext) {
...
+
+  const documentOpenHandler = vscode.window.onDidChangeActiveTextEditor(async () => {
```
DavisVaughan, line 139: I'm tempted to suggest that we move both of these listeners to `VisualEditorProvider.register()`. It gets called by `activateEditor()` above so it would still run at the right time, but:
- Both of these are about the editor, so putting it in `activateEditor()` makes sense to me
- In `register()` we already do things like register listeners for `workspace.onDidChangeTextDocument()`, so to me these are just "one more" editor related listeners and they belong beside the others.

apps/vscode/package.json
```diff
@@ -1036,6 +1036,16 @@
           "default": true,
           "markdownDescription": "When using a venv or conda environment, prefer Quarto CLI installed with pip in that environment. This will override Quarto CLI in the `PATH`, but not an explicitly configured `#quarto.path#`."
         },
+        "quarto.defaultEditor": {
+          "order": 11,
+          "type": "string",
+          "markdownDescription": "Default editor",
```
DavisVaughan, line 1042:
```suggestion
          "markdownDescription": "Default editor. Only applicable in Positron.",
```

This will be user facing in VS Code too so I think we need to say _something_ unless we can figure out how to make it work there too

apps/vscode/src/providers/editor/toggle.ts
```diff
@@ -12,16 +12,88 @@
...
+import * as path from 'path';
+import * as quarto from "quarto-core";
+import fs from "node:fs";
+import yaml from "js-yaml";
 import { Command } from "../../core/command";
 import { isQuartoDoc, kQuartoLanguageId } from "../../core/doc";
 import { VisualEditorProvider } from "./editor";
 
 
+export async function determineMode(doc: string): Promise<string | undefined> {
```
DavisVaughan, line 26: No longer async i think! Also update the call site!

apps/vscode/src/providers/editor/toggle.ts
```diff
@@ -12,16 +12,88 @@
...
+    editorOpener = workspaceHasQuartoYaml();
+  }
+
+  return editorOpener;
+}
 
+export async function setEditorOpener() {
+  const config = vscode.workspace.getConfiguration('quarto').get<string>('defaultEditor');
+  const viewType = config === 'visual' ? VisualEditorProvider.viewType : 'textEditor';
+  vscode.workspace.getConfiguration('workbench').update('editor.defaultView', viewType, true);
```
DavisVaughan, line 47: Is this line doing anything? What is `editor.defaultView` and why are we updating it?

apps/vscode/src/providers/editor/toggle.ts
```diff
@@ -12,16 +12,88 @@
...
+  await vscode.commands.executeCommand("workbench.action.setDefaultEditor",
+    vscode.Uri.file('filename.qmd'),
```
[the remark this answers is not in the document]
DavisVaughan: Actually I wonder if we could expose `updateUserAssociations()` with the slightly better name of `setEditorAssociations()`, but then expose it from `positron.d.ts` rather than as a command (either in the top level `'positron'` namespace, or under a new `'workbench'` sub namespace, mirroring what VS Code has. That would work like our existing `'language'` and `'runtime'` sub namespaces).

That would probably be nice because it would force you to look at Quarto's `hasHooks()` and `hooksApi()` helpers to determine whether or not you are in positron (we use that in a few other places too)

(FWIW `saveAllTitled` had to be a command to be able to call it from R, but this feels more like an Extension API feature that can be a "real" function exported from `positron.d.ts`)

apps/vscode/src/providers/editor/toggle.ts
```diff
@@ -12,16 +12,88 @@
...
+
+export function hasEditorMode(doc: string, mode: string) {
+  if (doc) {
+    const match = doc.match(quarto.kRegExYAML);
+    if (match) {
+      const yaml = match[0];
+      return (
+        !!yaml.match(new RegExp("^editor:\\s+" + mode + "\\s*$", "gm")) ||
+        !!yaml.match(new RegExp("^[ \\t]*" + mode + ":\\s*(default)?\\s*$", "gm"))
+      );
```
DavisVaughan, line 90: Oh I just remembered there is another way to specify source vs visual in the yaml that applies when you specify extra options, so we need to handle that too

Screenshot 2024-10-22 at 4 35 24 PM

apps/vscode/src/main.ts
```diff
@@ -125,9 +126,44 @@ export async function activate(context: vscode.ExtensionContext) {
...
+    // Check if the document language is "quarto"
+
```
DavisVaughan, line 141: It seems like we don't have things quite right in this changing active text editor listener

1. A diff editor view of a quarto document immediately closes


https://github.com/user-attachments/assets/05a10155-8061-43b6-a97e-e970ea5ee6eb



2. I can't seem to go from Visual Mode to Source Mode explicitly

https://github.com/user-attachments/assets/bd14536d-08a1-48d0-8f62-248ab2b05280

apps/vscode/src/providers/editor/toggle.ts
cscheid: We should really offer better facilities for editors to verify metadata. This works well enough for now, but it could come from a `_metadata.yml` file a couple of directories up, or a `_quarto.yml` configuration, or ...

cscheid: `quarto inspect` probably works, but I bet it's too slow to be usable without a large degree of annoyance.

apps/vscode/src/main.ts
isabelizimm: This makes a lot of sense to me! Moved 👍

apps/vscode/src/providers/editor/toggle.ts
isabelizimm: Having some sort of toolkit around the wrangling all the places/configurations the yaml metadata might be set would be super helpful! I've never used `quarto inspect` before. Trying it locally seems to take multiple seconds for a single, simple file, so I would +1 the "too slow to be usable" idea 😅

apps/vscode/package.json
```diff
@@ -1382,7 +1392,7 @@
     "@types/semver": "^7.3.13",
     "@types/tmp": "^0.2.3",
     "@types/uuid": "^9.0.0",
-    "@types/vscode": "1.66.0",
+    "@types/vscode": "1.94.0",
```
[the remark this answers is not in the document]
isabelizimm: Since it is in dev dependencies, I was not under the impression it will force people into a new vscode version? 

For CI purposes I can back down to a lower version, like 1.75.0, but to differentiate between a diff ediotr we do need the `vscode.window.tabGroups` API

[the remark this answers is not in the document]
DavisVaughan: Oh, then also note that this is in that file

```
  "engines": {
    "vscode": "^1.75.0"
  },
```

stating that Quarto has compatibility with vscode 1.75.0 and up. But if we unconditionally use types from a newer vs code, then that doesn't hold true

I'm not sure what happens if we try `window.tabGroups.activeTabGroup.activeTab?.label` when `window.tabGroups` doesn't exist

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -140,18 +150,38 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
...
       const editor = window.activeTextEditor;
-      if (!editor) {
-        return;
-      }
-      const document = editor.document;
-      if (document && isQuartoDoc(document)) {
-        const uri = document.uri.toString();
+      const visualEditor = VisualEditorProvider.activeEditor();
+
+      let document = editor?.document || visualEditor?.document;
```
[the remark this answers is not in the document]
isabelizimm: unfortunately, visual editors are not considered editors, so when we get the `onDidChangeActiveTextEditor` notification when opening a visual editor, the `editor` value is always `undefined`. I pulled the lines you're highlighting in the latest commit, since it is not providing anything useful.

This gets at the one bit that is a bit awkward: if the default is to open in a visual editor, this won't ever fire since vscode doesn't think there is ever an active editor. So if people have the default as source but put visual in a `*.qmd` file or `_quarto.yml` file, all will act as expected. BUT if you update the editor associations to visual, files will always open in the visual editor (you can still swap to source manually) since it doesn't fire `onDidChangeActiveEditor` changing between files opened in visual editors. 

An option to avoid this behavior is to use `onDidOpenTextDocument`, but then closing editors is unpredictable since there is only the ability to close active editors, not by name. I sort of think the current setup this is okay, since my guess is most people are setting editor mode in the qmd or `_quarto.yml` file?

juliasilge:
I got the `.vsix` from [here](https://github.com/quarto-dev/quarto/actions/runs/11576765507) to test out the behavior and unfortunately 😬 I see:

> command 'quarto.editInVisualMode' not found

Might something have gone wrong with how the command is registered?

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,50 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
       }
     }));
 
+    context.subscriptions.push(window.tabGroups.onDidChangeTabs(async (t) => {
+      if (t.closed.length > 0) { return; }
+      const tabs = t.changed.length > 0 ? t.changed : t.opened;
```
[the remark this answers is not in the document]
isabelizimm: Hmmm... I think we can just target opened tabs, and ignore tab changes. We will presume that the editor should stay the same as the user leaves it, eg, if you open something in source mode, go to another tab, update a `_quarto.yml` to have a default as visual, then going back to the original file should still be source (unless they close and reopen it).

`opened` tabs are never considered active, but I think if we're only looking at newly opened tabs, we can always treat them as if they are. That should eliminate the switching error you're seeing as well.

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,50 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
       }
     }));
 
+    context.subscriptions.push(window.tabGroups.onDidChangeTabs(async (t) => {
```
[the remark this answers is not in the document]
isabelizimm: I'm not sure if we would be able to move everything super cleanly; `Tab`s don't have access to a `TextEditor` AFAICT, which is how the user is navigated to the location

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,50 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
...
+
+            const isTextEditor = tab.input instanceof TabInputText;
+            const viewType = isTextEditor ? "textEditor" : tab.input.viewType;
+
+            // get file contents
+            const fileData = await workspace.fs.readFile(uri);
+            const fileContent = Buffer.from(fileData).toString('utf8');
+            const editorMode = determineMode(fileContent, uri);
+            let isSwitch = this.visualEditorPendingSwitchToSource.has(uri.toString()) || this.editorPendingSwitchToVisual.has(uri.toString());
+            this.editorPendingSwitchToVisual.delete(uri.toString());
```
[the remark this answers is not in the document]
isabelizimm: I'm not sure if we would be able to move everything super cleanly; `Tab`s don't have access to a `TextEditor` AFAICT, which is how the user is navigated to the location

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,50 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
...
+              const allTabs = window.tabGroups.all?.[0]?.tabs;
+
+              // find tab to close if swapping editor type. we don't want to close an active
+              // tab since a tab we are opening has not been set as active yet. we also don't
+              // want to close preview tabs since they will automatically be overriden
+              const tabsToClose = allTabs.filter(tab => (
+                tab.input?.uri?.toString() === uri?.toString()) && (tab.isActive == false || tab.isPreview == true)
+              );
+
+              await window.tabGroups.close(tabsToClose, false);
```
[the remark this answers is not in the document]
isabelizimm: I can't see the difference when testing empirically, but that makes sense to me 😆

apps/vscode/src/providers/editor/editor.ts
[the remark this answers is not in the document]
DavisVaughan: I played around with this for awhile and i think the problem is going to be that at `open` time the tabs aren't in the `window.tabGroup` array yet. You have to wait for them to become `isActive` first, which would come in a `change` event.

apps/vscode/src/providers/editor/editor.ts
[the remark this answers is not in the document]
DavisVaughan: ah good point

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,52 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
...
+            const fileData = await workspace.fs.readFile(uri);
+            const fileContent = Buffer.from(fileData).toString('utf8');
+            const editorMode = determineMode(fileContent, uri);
+            let isSwitch = this.visualEditorPendingSwitchToSource.has(uri.toString()) || this.editorPendingSwitchToVisual.has(uri.toString());
+            if (this.editorPendingSwitchToVisual.has(uri.toString())) {
+              this.editorPendingSwitchToVisual.delete(uri.toString());
+            }
+
+            if (editorMode && editorMode != viewType && !isSwitch) {
+              const allTabs = window.tabGroups.all?.[0]?.tabs;
```
DavisVaughan, line 168: I think we need a big meaningful comment here, because this is (at first glance) _really_ weird that we have to do this.

In my experience playing around with this if you set `await window.tabGroups.close(tab, true);` using the `tab` you get from the change event, it won't close and you get a "tab does not exist" error in the debug console.

From what I can tell, the `tab` we get in the change event is not PRECISELY equal to the tabs that exist in `window.tabGroups` at this point in time. It's possible that during this `opened` event they are a little out of sync. The tab we care about closing does exist in there, we just have to find it by matching by URI - something we know is stable across `change` events.

So I think it looks like this:

```
// Unfortunately we cannot just `await window.tabGroups.close(tab, true);`.
// The `tab` we get from the change event seems to not be precisely the same
// as the tab in `window.tabGroups`, so if we try and close `tab` we get a "tab
// not found" error. The one we care about does exist in `window.tabGroups`,
// we just have to manually find it using the URI as a stable field to match on.
```

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,52 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
...
+
+              // find tab to close if swapping editor type. we don't want to close an active
+              // tab since a tab we are opening has not been set as active yet. we also don't
+              // want to close preview tabs since they will automatically be overriden
+              const tabsToClose = allTabs.filter(tab =>
```
DavisVaughan, line 173: I believe we are expecting exactly 1 tab to match here, so you could use `find()` instead of `filter()` and only do the close/open dance when we `find()` a tab, like

```ts
const tabToClose = allTabs.find(stuffhere);

if (tabToClose) {
  await window.tabGroups.close(tabToClose, true);
  await commands.executeCommand("vscode.openWith", uri, editorMode);
}
```

DavisVaughan: That might also be _slightly_ safer in some weird edge cases. Like with the "double open" bugs I've reported in this review, we would not have actually opened the visual editor if we programmed it this way, because `tabToClose` would have been `undefined` and the entire branch would not have run (it still would have been a bug on our part, just a slightly less bad one).

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,52 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
       }
     }));
 
+    context.subscriptions.push(window.tabGroups.onDidChangeTabs(async (t) => {
+      const tabs = t.opened;
+
+      if (tabs.length > 0) {
```
DavisVaughan, line 149: can still flatten this if you want

DavisVaughan (approved):
A few more small bugs but looks good enough to approve now - the timing of all this stuff is so confusing!

Don't forget a CHANGELOG bullet!

apps/vscode/src/providers/editor/editor.ts
```diff
@@ -139,8 +143,52 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
...
+            if (editorMode && editorMode != viewType && !isSwitch) {
+              const allTabs = window.tabGroups.all?.[0]?.tabs;
+
+              // find tab to close if swapping editor type. we don't want to close an active
+              // tab since a tab we are opening has not been set as active yet. we also don't
+              // want to close preview tabs since they will automatically be overriden
+              const tabsToClose = allTabs.filter(tab =>
+                ((tab.input instanceof TabInputText) || (tab.input instanceof TabInputCustom)) &&
+                (tab.input?.uri?.toString() === uri?.toString()) &&
+                (tab.isActive == false || tab.isPreview == true)
```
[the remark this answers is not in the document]
isabelizimm: it does sort of break without checking for preview. in this scenario, my default editor is the text editor. if i switch between a document specifying source vs visual, i end up with double editors for the document that should be visual (where one editor is source mode and one editor is visual mode)

https://github.com/user-attachments/assets/888ddd49-8790-49d8-b79d-5f18be65750c

PR #585, opened 2024-10-24
Try using a unique temp file for vdocs

juliasilge:
_Maybe_ addresses a category of problem some users are experiencing in Positron:

- https://github.com/posit-dev/positron/issues/3792
- https://github.com/posit-dev/positron/issues/3945
- https://github.com/posit-dev/positron/discussions/4437
- https://github.com/posit-dev/positron-beta/discussions/227

No one on the team has been able to observe this problem directly but it is clearly real. @DavisVaughan came up with the hypothesis that maybe the problem is that the `vdoc` tempfiles are getting overwritten by different requests. Quoted from Slack:

> whenever an LSP feature is fired that requires a virtual doc, quarto takes the `TextDocument`, turns it into virtual doc form (with comments in place of prose and code block fences), and writes it to a temp file
> then that temp file is used by the lsp, and it is deleted on the next lsp request
> here's where i think the problem may be. every single one of these temp files for R gets the same name, `.intellisense.r`
> 
> since we are in async world here is what may be happening:
> - a hover request starts with a particular position, we make the virtual doc, `.intellisense.r` (hover), but then hit an `await` before we complete it
> - maybe some text edits happen
> - a completion request immediately fires, we start handling that. to make its virtual doc we overwrite the original one
> - we get back to the hover request after the `await`, but now the position doesn't line up with the contents of the `.intellisense.r` document
> - ark finally handles it and things blow up

This PR uses UUIDs to make a whole bunch of different tempfiles, one for each request, rather than one file that could get deleted before the request can be handled. This is maybe overkill but we can have folks get the `.vsix` to try it out.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,7 +6,7 @@ on:
     branches:
       - main
     paths:
-      - 'apps/vscode/**' 
+      - 'apps/vscode/**'
   pull_request:
     branches:
       - main
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,12 +6,12 @@ on:
     branches:
       - main
     paths:
-      - apps/vscode 
+      - 'apps/vscode/**' 
   pull_request:
     branches:
       - main
     paths:
-      - apps/vscode 
+      - 'apps/vscode/**' 
   workflow_dispatch:
 
 jobs:
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.117.0 (unreleased)
 
+- Fix issue with temp files for LSP request virtual documents (<https://github.com/quarto-dev/quarto/pull/585>)
+
 ## 1.116.0 (Release on 2024-10-08)
 
 - Fix issue with raw html blocks being removed from document by Visual Editor (<https://github.com/quarto-dev/quarto/issues/552>)
--- a/apps/vscode/src/lsp/client.ts
+++ b/apps/vscode/src/lsp/client.ts
@@ -53,7 +53,6 @@ import {
   virtualDocUri,
 } from "../vdoc/vdoc";
 import { activateVirtualDocEmbeddedContent } from "../vdoc/vdoc-content";
-import { deactivateVirtualDocTempFiles, isLanguageVirtualDoc } from "../vdoc/vdoc-tempfile";
 import { vdocCompletions } from "../vdoc/vdoc-completion";
 
 import {
@@ -160,8 +159,6 @@ export async function activateLsp(
 }
 
 export function deactivate(): Thenable<void> | undefined {
-  deactivateVirtualDocTempFiles();
-
   if (!client) {
     return undefined;
   }
@@ -289,8 +286,7 @@ function embeddedGoToDefinitionProvider(engine: MarkdownEngine) {
           adjustedPosition(vdoc.language, position)
         );
         const resolveLocation = (location: Location) => {
-          if (isLanguageVirtualDoc(vdoc.language, location.uri) ||
-            location.uri.toString() === vdocUri.uri.toString()) {
+          if (location.uri.toString() === vdocUri.uri.toString()) {
             return new Location(
               document.uri,
               unadjustedRange(vdoc.language, location.range)
@@ -300,8 +296,7 @@ function embeddedGoToDefinitionProvider(engine: MarkdownEngine) {
           }
         };
         const resolveLocationLink = (location: LocationLink) => {
-          if (isLanguageVirtualDoc(vdoc.language, location.targetUri) ||
-            location.targetUri.toString() === vdocUri.uri.toString()) {
+          if (location.targetUri.toString() === vdocUri.uri.toString()) {
             const locationLink: LocationLink = {
               targetRange: unadjustedRange(vdoc.language, location.targetRange),
               originSelectionRange: location.originSelectionRange
@@ -349,5 +344,3 @@ function isWithinYamlComment(doc: TextDocument, pos: Position) {
   const line = doc.lineAt(pos.line).text;
   return !!line.match(/^\s*#\s*\| /);
 }
-
-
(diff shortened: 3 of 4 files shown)


cscheid:
Oof, that's a nasty failure case, thanks for finding it!

I like the uuid solution a lot, and this is good to merge as far as I'm concerned.

One question for our collective pondering: how concerned are we about creating large numbers of temporary files here? If we're getting multiple `virtualDocUriFromTempFile` calls that race each other, should we also be concerned about leaking files? Like I said, I think this is good to merge as is, because at worst it trades a crashing bug for a resource leak and that's a win. But maybe we can fix that too?

juliasilge:
Looks like success 🤞 reported in https://github.com/posit-dev/positron/issues/3792#issuecomment-2436568588. I've pointed the other folks here to get the `.vsix` as well so let's see if folks are able to try it out.

juliasilge:
For the long term, is it possible to use in-memory [virtual documents](https://code.visualstudio.com/api/extension-guides/virtual-documents)? Like with a true LSP backend? Do you have context for why the extension is using temp files in this way?

In the short term, I definitely would advocate for merging something along the lines of this PR, pending another person or two confirming that this fixes their problem, and releasing the extension ASAP. This has been probably _the_ most pernicious crash type Positron users have run into since we went public!

juliasilge:
Looks like another report of success here: https://github.com/posit-dev/positron/issues/3945#issuecomment-2437762858

cscheid:
> Do you have context for why the extension is using temp files in this way?

I don't. I wasn't involved in writing that code, so I'm reading it for the first time and figuring it out.

It looks like `virtualDocUri` in `apps/vscode/src/vdoc/vdoc.ts` calls `virtualDocUriFromTempFile` if `virtualDoc.language.type === "tempfile"`, where `EmbeddedLanguage.type` is `"content" | "tempfile"`. In `apps/vscode/src/vdoc/languages.ts`, there's an additional hint at kEmbededLanguages (sic) that mentions languages requiring creating a temp file. The last four languages declared have `{type: "content"}` and a comment about "working with text document content provider", so I assume the other ones didn't work for some reason and requires the tempfile route.

DavisVaughan:
> should we also be concerned about leaking files

We definitely leak files like crazy now, since we never clean up `.intellisense.*.R` now and we make one at each LSP call (so, A LOT)

DavisVaughan:
Oh wait, I'm dumb. These are all LSP _requests_. Meaning we get a _reply_ back from the server when its done. Duh! So when we get the reply we just clean up the temp file. So if we want to keep the tempfile approach I think we can make this work without leaking files.

We are even doing this already! In this `local` branch the tempfile we create has a `cleanup` hook attached that deletes it
https://github.com/quarto-dev/quarto/blob/97cc659a3481aca86b04e6c3fb5015d204d83ee8/apps/vscode/src/vdoc/vdoc-tempfile.ts#L51

And we do call this cleanup hook in a number of places. Here is an example:
https://github.com/quarto-dev/quarto/blob/97cc659a3481aca86b04e6c3fb5015d204d83ee8/apps/vscode/src/lsp/client.ts#L237

But what I really like is this `withVirtualDocUri()` helper that keeps you from footgunning yourself
https://github.com/quarto-dev/quarto/blob/97cc659a3481aca86b04e6c3fb5015d204d83ee8/apps/vscode/src/vdoc/vdoc.ts#L124-L132

We should consider using this everywhere we use a virtual doc right now, and consider taking it a step further as just `withVirtualDoc()` to include the whole virtual doc creation / teardown, rather than just the URI part.

DavisVaughan:
Regarding _why_ we do the tempfile approach. It is possible that some LSP requests require _workspace context_. So by putting the tempfile in the workspace alongside the qmd (where other R files might also exist), then the LSP has the capabilities to process the request on the tempfile _with the context of other R files around it_

lionel-:
(Sorry I pushed a commit here by mistake, I force-pushed it away)

lionel-:
Some more observations:

As long as a document is opened (in the sense of `vscode.workspace.openTextDocument()`), all changes are tracked by VSCode and corresponding events emitted in the correct order. If a document is closed, VSCode stops tracking changes. In both cases (document changed and document closed) code assistance requests should be cancelled.

Cancellation works via a cancellation token that must be checked manually by assistance providers. We currently fail to do that, and instead propagate the assistance request to LSP backends. These backends then are at risk of operating on outdated documents and provide incorrect results or even crash.

The workaround taken in this PR is reasonable. It creates a unique document for each request and any further change causes a new unique document to be created. This guarantees that requests operate on an unchanging source and prevents races. That may be the safest way we can implement the bridge between chunks and assistance providers. And if the provider returns results for outdated documents, VSCode will ignore them. So as long as it doesn't cause performance issues, we probably want to keep doing that.

However this approach only hides the problem. We should also make an effort to cancel propagation of assistance requests to LSPs, if only to avoid unnecessary traffic with the backends. Here is an example with the completion provider: https://github.com/quarto-dev/quarto/commit/99be12411ea5a28d080879e64dfc86acc05d72f1

> We should consider using this everywhere we use a virtual doc right now, and consider taking it a step further as just withVirtualDoc() to include the whole virtual doc creation / teardown, rather than just the URI part.

I was looking at this too, should be easy enough: https://github.com/quarto-dev/quarto/commit/0eaea034cdbde0cabe213522fc81699ca9b6a670


> Regarding why we do the tempfile approach. It is possible that some LSP requests require workspace context. So by putting the tempfile in the workspace alongside the qmd (where other R files might also exist), then the LSP has the capabilities to process the request on the tempfile with the context of other R files around it

IIUC this is the "local" code path, currently used for formatting and jump-to-definition. This stores the files alongside the qmd files. But here we are working with the non-local path that stores files in a temp folder. I _think_ files from a virtual document provider would be treated in the same way, and be part of the workspace just like these tempfiles?

lionel- (commented):
@cscheid If you have time would you mind doing a final review? If not no worries, we've closely looked at it with @DavisVaughan already.

We're aiming to merge before Wednesday end of day so this can get in the next Positron release.

juliasilge:
Another report of success here: https://github.com/posit-dev/positron/discussions/4437#discussioncomment-11089524

juliasilge:
Carlos is OOO so let's go with the work as is, thoroughly checked out by Lionel and Davis and confirmed by our couple of Positron users who were having trouble. ✅

PR #589, opened 2024-10-28
Use Positron's statement range provider for Python docs, when available

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/1646

You'll notice this is the first time these files have been edited since the formatting was added to this repo; I vote for just muddling through the unrelated formatting changes, if that's OK.

There are two changes needed to get `quarto.runCurrent` to work as we want it to in a Python `.qmd`:

- Before this PR, that command resolved to the command for "run cell" so we need to not do that anymore, since this is a situation where we can identify a statement to run.
- When I worked on https://github.com/quarto-dev/quarto/pull/436, I did not correctly use `unadjustedRange()` on the way back out of the statement range provider, because R doesn't have anything injected. Turns out that Python does:

https://github.com/quarto-dev/quarto/blob/2549e525e100ae4e0cb619be9dbbac7f29db98b8/apps/vscode/src/vdoc/languages.ts#L43

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.117.0 (unreleased)
 
+- Improved statement execution for Python `.qmd` files in Positron (<https://github.com/quarto-dev/quarto/pull/589>)
+
 ## 1.116.0 (Release on 2024-10-08)
 
 - Fix issue with raw html blocks being removed from document by Visual Editor (<https://github.com/quarto-dev/quarto/issues/552>)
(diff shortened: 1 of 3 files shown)


juliasilge:
You [can get the `.vsix` from here](https://github.com/quarto-dev/quarto/actions/runs/11555962603) to test this out, and what you'll see is that <kbd>Cmd</kbd>+<kbd>Enter</kbd> now steps through single lines in a Python `.qmd`, like it has for a while for R. There are some slight differences in behavior between R and Python when you are at the end of a chunk, which we could address with https://github.com/posit-dev/positron/issues/3233.

cscheid (approved):
👍

PR #593, opened 2024-10-30
Don't minify during development

DavisVaughan:
Working in `apps/vscode/` currently kind of sucks because the debugging experience is suboptimal. Sourcemaps don't seem to be working right, causing the Debug Variables pane to show minified names, not real names. Stepping is also pretty jumpy.

Turns out that we are minifying during our `dev` builds, which seems to be what is breaking things. It is also kind of weird right? We should only minify during _production_, not during development. That's also what VS Code's extension guide suggests:
https://code.visualstudio.com/api/working-with-extensions/bundling-extension#run-esbuild

And what esbuild's guide says: "Usually you minify code in production but not in development."
https://esbuild.github.io/api/#minify

Before:

(Note how the variables pane has 1 letter names, and how it jumps over the if branch, seemingly without running it)

https://github.com/user-attachments/assets/de898ddd-a5b9-4c14-b5f2-c0941c3e293e

After:

https://github.com/user-attachments/assets/59a13121-b4d9-461d-9b22-6be7ace87c4a

---

In theory both VS Code and esbuild have all the required glue to make `minify + sourcemaps` work through this `names` mapping field, but clearly something isn't working right:
https://github.com/evanw/esbuild/issues/1296
https://github.com/microsoft/vscode/issues/12066

### DIFF
--- a/packages/build/src/index.ts
+++ b/packages/build/src/index.ts
@@ -21,7 +21,6 @@ export interface BuildOptions {
   outfile: string;
   assets?: Array<AssetPair>;
   bundle?: boolean;    // true
-  minify?: boolean;    // false
   format?: Format;     // cjs
   platform?: Platform; // node
   external?: string[]; // []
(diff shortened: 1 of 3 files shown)


packages/build/src/index.ts
```diff
@@ -36,25 +35,24 @@ export async function runBuild(options: BuildOptions) {
...
     dev = false
   } = options;
 
-  await build({ 
+  await build({
     entryPoints,
     outfile,
     bundle,
-    minify,
+    minify: !dev,
```
DavisVaughan, line 46: We now:
- Always minify during production
- Never minify during development

I don't think we were minifying during production before (we were only doing it during dev, oops).

Is this `build()` hook what is also used when generating the production release? If so, this is technically a change in what we'd ship, but probably a good one?

cscheid (commented):
Do we want to split the difference and keep the minify option, but defer to dev if minify is undefined? I'm not sure I want to remove minify before understanding exactly how it could be used.

DavisVaughan:
I've kept the exposed `minify` option and instead just set it to `minify: !dev,` in the two places that we call `runBuild()`.

Issue #580, opened 2024-10-21
Positron's help panel is not used when F1 is invoked


andrewheiss:
Currently, in VS Code, pressing F1 inside a function name opens the documentation in a special "Quarto: Help" panel in the File Explorer:

![Image](https://github.com/user-attachments/assets/dff8eba0-9c60-490f-b3a3-3543f3c5da72)

That's great when using VS Code because it doesn't have a dedicated help panel like RStudio does.

However, in the spirit of RStudio, Positron *does* have a dedicated help panel—when pressing F1 in a regular R script in Positron, the documentation opens there:

![Image](https://github.com/user-attachments/assets/a05c5a65-cd02-4b5d-8608-65a0e8208162)

When pressing F1 in a Quarto file in Positron, though, nothing happens in the Positron help panel. Instead, the documentation appears in the custom Quarto help panel:

![Image](https://github.com/user-attachments/assets/457d3ddf-dcda-4014-836c-b7421f61d2ff)

There's an issue at Positron tracking this: https://github.com/posit-dev/positron/issues/4097 and it's part of a larger push for better Quarto improvements (https://github.com/posit-dev/positron/issues/4858). I figured that F1 wasn't doing anything at all (I'd forgotten about Quarto's custom help panel), and others likely thought the same, but a recent comment https://github.com/posit-dev/positron/issues/4097#issuecomment-2427178059 discovered that F1 is actually working as expected in Positron—it's just invisible if the custom Quarto help panel is hidden.

So the underlying issue seems to be that Quarto's VS Code extension is hijacking F1 inside Positron

juliasilge:
It turns out this is the `quarto-assist` webview the Quarto extension contributes; it can be pinned/unpinned with commands `quarto.assist.pin` and `quarto.assist.unpin`. There are three kinds of things that can get rendered in this webview:

https://github.com/quarto-dev/quarto/blob/227af7814a1e42737383ed4ce361242c36504dad/apps/vscode/src/providers/assist/render-assist.ts#L43-L45

Once you get this webview going, it responds to where your cursor is and renders things for you, depending on whether it is an equation, a function etc. The Help does not seem to work for Python for me, and I have not found a pydoc server or anything like that.

One option would be to turn this webview off in Positron and hook up <kbd>F1</kbd> to calling Positron's Help; it does seem like most use of it is for Help from the issues about it. Or I guess we can pipe through <kbd>F1</kbd> to Positron's Help without messing with this at all, and they both would be available.

[...]

PR #599, opened 2024-11-06
Add Positron help topic provider

juliasilge:
Addresses https://github.com/quarto-dev/quarto/issues/580 as well as https://github.com/posit-dev/positron/issues/4097

Goes together with https://github.com/posit-dev/positron/pull/5292 which is what will power this

I did _not_ end up messing with the `quarto-assist` webview at all, since that is a whole separate thing; this is just about sending info to Positron for the Positron Help panel.

### DIFF
--- a/apps/vscode/src/host/hooks.ts
+++ b/apps/vscode/src/host/hooks.ts
@@ -16,7 +16,7 @@
 import * as vscode from 'vscode';
 import * as hooks from 'positron';
 
-import { ExtensionHost, HostWebviewPanel, HostStatementRangeProvider } from '.';
+import { ExtensionHost, HostWebviewPanel, HostStatementRangeProvider, HostHelpTopicProvider } from '.';
 import { CellExecutor, cellExecutorForLanguage, executableLanguages, isKnitrDocument, pythonWithReticulate } from './executors';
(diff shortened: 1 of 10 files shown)


apps/vscode/src/host/hooks.ts
```diff
@@ -186,3 +195,39 @@ async function getStatementRange(
...
+    readonly engine: MarkdownEngine,
+  ) {
+    this._engine = engine;
+  }
+
+  async provideHelpTopic(
+    document: vscode.TextDocument,
+    position: vscode.Position,
+    token: vscode.CancellationToken): Promise<string | undefined> {
+    const vdoc = await virtualDoc(document, position, this._engine);
```
juliasilge, line 212: There is a lot of boilerplate here for the `virtualDoc` if you compare this to what we do for `provideStatementRange()`; it might be interesting in the longer term to see if we can do anything to abstract that away.

DavisVaughan (approved):
Looks nice! Approved but also check out https://github.com/quarto-dev/quarto/pull/601 before merging (feel free to merge that into this if you agree with my PR over there)


https://github.com/user-attachments/assets/8a34efd1-7a5c-4865-beb1-9862e87c024f

PR #601, opened 2024-11-08
Help topic provider with doc

DavisVaughan:
Branched from https://github.com/quarto-dev/quarto/pull/599

@juliasilge what do you think about this?

`withVirtualDocUri()` is now in charge of _both_:
- Creation of the vdocUri by calling `virtualDocUri()`
- Clean up of the vdocUri

This prevents any kind of leakage of temp files from occurring. 

After this we should try and go back and use `withVirtualDocUri()` instead of `virtualDocUri()` everywhere and then make `virtualDocUri()` private.

`withVirtualDocUri()` could also provide an error hook if a user needs to control the `} catch (error) {` case too, but I'd only add that later on if it turns out we need it.

### DIFF
--- a/apps/vscode/src/core/hover.ts
+++ b/apps/vscode/src/core/hover.ts
@@ -15,17 +15,18 @@
 
 
 import { Hover, MarkdownString, MarkedString, Position, SignatureHelp, commands } from "vscode";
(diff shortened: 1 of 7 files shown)


DavisVaughan:
CC @lionel- too, just as an FYI since we've talked about this in the past a bit

juliasilge (approved):
Nice! That is going to work well.

PR #606, opened 2024-11-21
Add C# cell executor support in VSCode

momr:
Add C# cell executor support in VSCode similar to Python and R

image


cscheid:
This looks good! We will need a contributor agreement so we can do a copyright transfer, though. Can you follow the instructions [here](https://github.com/quarto-dev/quarto-cli/blob/main/CONTRIBUTING.md) and let us know when it's done?

Thanks!

PR #608, opened 2024-11-22
Set context keys for document language

juliasilge:
This is a revamp of the work in #607.

Unfortunately after I started trying to use this in Positron, it just wasn't early enough or quick enough to set these context keys in `languageAtPosition()`. I believe that only gets called if we are trying to do one of these:

https://github.com/quarto-dev/quarto/blob/d9d15151187cd2222b0ce47e1654f3a92031b177/apps/vscode/src/vdoc/vdoc.ts#L114-L121

It's just not enough, especially if you are switching back and forth between different files; it's very easy to get in a state where you end up with R shortcuts in Python chunks and there are lots of kinds of edits I could make that did not trigger one of these.

New idea! This PR tries out setting these context keys when we paint the background. In my experience so far, this feels a lot better, and we already are identifying all the `tokens` so not much extra work.

### DIFF
--- a/apps/vscode/src/providers/background.ts
+++ b/apps/vscode/src/providers/background.ts
@@ -22,6 +22,7 @@ import { isQuartoDoc, kQuartoDocSelector } from "../core/doc";
 import { MarkdownEngine } from "../markdown/engine";
 import { isExecutableLanguageBlock } from "quarto-core";
 import { vscRange } from "../core/range";
+import { mainLanguage } from "../vdoc/vdoc";
 
 export function activateBackgroundHighlighter(
   context: vscode.ExtensionContext,
@@ -161,6 +162,10 @@ async function setEditorHighlightDecorations(
     for (const block of tokens.filter(isExecutableLanguageBlock)) {
       blockRanges.push(vscRange(block.range));
     }
+
+    // expose cell language for use in keybindings, etc
+    const language = mainLanguage(tokens);
+    vscode.commands.executeCommand('setContext', 'quartoLangId', language?.ids[0]);
 
     // find inline executable code
     for (let i = 0; i < editor.document.lineCount; i++) {
--- a/apps/vscode/src/providers/background.ts
+++ b/apps/vscode/src/providers/background.ts
@@ -22,7 +22,6 @@ import { isQuartoDoc, kQuartoDocSelector } from "../core/doc";
 import { MarkdownEngine } from "../markdown/engine";
 import { isExecutableLanguageBlock } from "quarto-core";
 import { vscRange } from "../core/range";
-import { mainLanguage } from "../vdoc/vdoc";
 
 export function activateBackgroundHighlighter(
   context: vscode.ExtensionContext,
@@ -162,10 +161,6 @@ async function setEditorHighlightDecorations(
     for (const block of tokens.filter(isExecutableLanguageBlock)) {
       blockRanges.push(vscRange(block.range));
     }
-
-    // expose cell language for use in keybindings, etc
-    const language = mainLanguage(tokens);
-    vscode.commands.executeCommand('setContext', 'quartoLangId', language?.ids[0]);
 
     // find inline executable code
     for (let i = 0; i < editor.document.lineCount; i++) {
--- a/apps/vscode/src/providers/context-keys.ts
+++ b/apps/vscode/src/providers/context-keys.ts
@@ -1,0 +1,93 @@
+/*
+ * context-keys.ts
+ *
+ * Copyright (C) 2024 by Posit Software, PBC
+ *
+ * Unless you have received this program directly from Posit Software pursuant
+ * to the terms of a commercial license agreement with Posit Software, then
+ * this program is licensed to you under the terms of version 3 of the
+ * GNU Affero General Public License. This program is distributed WITHOUT
+ * ANY EXPRESS OR IMPLIED WARRANTY, INCLUDING THOSE OF NON-INFRINGEMENT,
+ * MERCHANTABILITY OR FITNESS FOR A PARTICULAR PURPOSE. Please refer to the
+ * AGPL (http://www.gnu.org/licenses/agpl-3.0.txt) for more details.
+ *
+ */
+
+
+import * as vscode from "vscode";
+import debounce from "lodash.debounce";
+
+import { isQuartoDoc } from "../core/doc";
+import { MarkdownEngine } from "../markdown/engine";
+import { mainLanguage } from "../vdoc/vdoc";
+
+const debounceOnDidChangeDocumentMs = 250;
+
+export function activateContextKeySetter(
+  context: vscode.ExtensionContext,
+  engine: MarkdownEngine
+) {
+
+  // set context keys when docs are opened
+  vscode.workspace.onDidOpenTextDocument(
+    (doc) => {
+      if (doc === vscode.window.activeTextEditor?.document) {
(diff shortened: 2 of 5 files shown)


juliasilge:
Related to https://github.com/posit-dev/positron/pull/5451

apps/vscode/src/providers/background.ts
```diff
@@ -162,6 +163,10 @@ async function setEditorHighlightDecorations(
       blockRanges.push(vscRange(block.range));
```
DavisVaughan, line 163: What if `highlightingConfig` isn't enabled? See the if-block above.

apps/vscode/src/providers/background.ts
```diff
@@ -162,6 +163,10 @@ async function setEditorHighlightDecorations(
...
     }
 
+    // expose cell language for use in keybindings, etc
+    const language = mainLanguage(tokens);
+    vscode.commands.executeCommand('setContext', 'quartoLangId', language?.ids[0]);
+
```
DavisVaughan, line 169: I do not feel it is appropriate to set the context here

Currently this function is quite pure. It is _just_ about setting highlight decorations and is very easy to understand.

If we take a look at _why_ it works to put this here, its because this is called from

- `triggerUpdateAllEditorsDecorations()`, which is called during the `onDidChangeVisibleTextEditors()` hook
- `triggerUpdateActiveEditorDecorations()`, which is called during the `onDidOpenTextDocument()` hook

That hook registration is part of `activateBackgroundHighlighter()`, which is called in `activateCommon()`.

---

I think what we really want is:
- In `activateCommon()`,
- Add a new `activateLanguageContextSetter()`
- Which registers `onDidOpenTextDocument()` and `onDidChangeVisibleTextEditors()` hooks of its own, and those hooks set the `quartoMainLanguageId` of the main language for the document.

Yes you have to reparse the document, but I think that is worth it over muddying the waters about what `setEditorHighlightDecorations()` is about, plus it gives you finer control over when the context is updated.

apps/vscode/src/providers/background.ts
```diff
@@ -162,6 +163,10 @@ async function setEditorHighlightDecorations(
       blockRanges.push(vscRange(block.range));
     }
 
+    // expose cell language for use in keybindings, etc
+    const language = mainLanguage(tokens);
+    vscode.commands.executeCommand('setContext', 'quartoLangId', language?.ids[0]);
```
DavisVaughan, line 168: I'd consider using a non-abbreviated and very descriptive name

```
quarto.document.mainLanguageId
quartoDocumentMainLanguageId
```

See `"quarto.assistView.isEnabled"` as a nice example

It's a little tricky because we have a mix of styles right now for context key naming. Sometimes we do `quarto.` sometimes we dont https://github.com/search?q=repo%3Aquarto-dev%2Fquarto%20setContext&type=code

I tend to like using the `.` as a way to namespace it to quarto

apps/vscode/src/vdoc/vdoc.ts
```diff
@@ -157,12 +157,8 @@ export async function virtualDocUri(
 export function languageAtPosition(tokens: Token[], position: Position) {
   const block = languageBlockAtPosition(tokens, position);
   if (block) {
```
DavisVaughan, line 159: IIUC, previously it _attempted_ to update the context key dynamically on each cell "activation", right? That does seem hard to get 100% right.

And now it tries to update the context key 1 time for whatever the main language in the document is.

I suppose that _could_ be mildly confusing. If you have 1 python cell in a mostly R document, then clicking in the python cell and running a keybinding will run the R version. That's kinda confusing, but I'm not sure what we can do better.

DavisVaughan (changes requested):

apps/vscode/src/vdoc/vdoc.ts
juliasilge: I did hope for us to get this working really nicely for reticulate Quarto files (per cell), but when I tried to use the approach with `languageAtPosition()` in practice, it felt quite awful. So yep, I think for now this approach with the "main language" is the best option.

apps/vscode/src/providers/background.ts
cscheid: Why not just `quarto.document.languageId` vs (eventually) something like `quarto.cell.languageId`?

`document` vs `cell` is an evocative distinction. `main` wants to contrast with `secondary`, and that's not what you're setting up to do in the future.

apps/vscode/src/providers/background.ts
juliasilge: This is a great idea! Implemented in c6f28f8673d712ba7d430fd6ee48e2e38fb39970 and working really well in https://github.com/posit-dev/positron/pull/5451.

apps/vscode/src/providers/background.ts
DavisVaughan: Works great for me!

apps/vscode/src/providers/context-keys.ts
```diff
@@ -0,0 +1,93 @@
...
+  // set context keys at activation time
+  triggerUpdateContextKeys(engine);
+
+}
+
+function triggerUpdateContextKeys(engine: MarkdownEngine) {
+  for (const editor of vscode.window.visibleTextEditors) {
+    setContextKeys(editor, engine);
+  }
+}
```
DavisVaughan, line 79: I think this is looking great!

My main comment is about this, and any other usage of `visibleTextEditors`

IIUC, the context key is "global". i.e. there is _not_ a context key per document, there is exactly 1 context key set globally.

If that is the case, then the context key should probably only be set for the _active_ text editor, not for _all visible_ text editors (because only 1 wins!).

For example, if I open 2 qmds side by side in split panes, one that is mostly R and one that is mostly Python, then I'm not entirely sure which is going to win here because both are "visible". However, only one at a time is ever "active", and I feel like that is probably what we should be targeting with this context key.

Is it possible all we really need to do is having a single hook for `onDidChangeActiveTextEditor()`?

Does that make any sense?

DavisVaughan (approved):
Looks great, just one main comment left about exactly when we should set the context key

apps/vscode/src/providers/context-keys.ts
juliasilge: Makes _total_ sense, yep. Changed in 8d00a51d44519ed9d4e47abdc47122e3c8e07151.

juliasilge: I added the hook for `onDidChangeTextDocument()` for the case where someone changes their document from being Python to R or vice versa. That is certainly rare, but it feels fairly broken if we don't include it.

juliasilge: You are correct that we don't need `onDidOpenTextDocument()` if we have `onDidChangeActiveTextEditor()`. 👍

juliasilge:
I installed the final `.vsix` from this PR and ran through all my normal Quarto moves, both with a current release build of Positron, dev build from https://github.com/posit-dev/positron/pull/5451, and VS Code. Behaves as expected and looks good!

apps/vscode/src/providers/context-keys.ts
DavisVaughan: > I added the hook for onDidChangeTextDocument()

Makes sense because you also dont have a "main" language at the start of a document until you actually start typing and adding some chunks (or maybe there is a default, but it likely isnt right)

---

PR looks great nice job!

PR #609, opened 2024-11-22
Do not force snippet suggestions to the bottom of completions

juliasilge:
Currently, we force snippet suggestions to the bottom of the completions list:

https://github.com/quarto-dev/quarto/blob/d818dd022d803ea93db34d1a6ccca56a05d54348/apps/vscode/package.json#L899

If we don't force this, then they will go wherever the underlying `embeddedCodeCompletionProvider` wants them to go. I'll install the new `.vsix` in VS Code and Positron to show where snippets show up with this proposed change.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Provide F1 help at cursor in Positron (<https://github.com/quarto-dev/quarto/pull/599>)
 - Expose new context keys for the language of a specific cell (<https://github.com/quarto-dev/quarto/pull/607>)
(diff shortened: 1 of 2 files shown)


juliasilge:
Before this PR, snippets were always pushed down to the bottom of the completions list. With this new change, the snippets go wherever the underlying completion provider puts them.

For example, in VS Code with vscode-R extension, the snippet for `lib` is set with the `snippetSortOrder` of "inline", which means alphabetically with everything else:

Screenshot 2024-11-22 at 3 43 31 PM

In Positron, the R snippets are set to show up with `snippetSortOrder` of "top":

Screenshot 2024-11-22 at 3 46 27 PM

Quarto probably shouldn't interfere with this ordering.

PR #644, opened 2025-01-20
fix: Use text area for alternative text input in image editor

mcanouil:
Replace the input field with a text area for entering alternative text in the image editing dialog, enhancing user experience.

### DIFF
--- a/packages/editor-ui/src/dialogs/edit-image.tsx
+++ b/packages/editor-ui/src/dialogs/edit-image.tsx
@@ -28,7 +28,8 @@ import {
   useId, 
   RadioGroup, 
   Label, 
-  Radio 
+  Radio,
+  Textarea
 } from "@fluentui/react-components";
 
 import { ModalDialog, ModalDialogTabList, showValueEditorDialog} from "ui-widgets";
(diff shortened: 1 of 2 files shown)


packages/editor-ui/src/dialogs/edit-image.tsx
```diff
@@ -162,10 +163,11 @@ const imagePanel =
   
     {alt !== undefined
       ?  <Field label={t("Alternative text")}>
-          <Input 
-            value={alt} 
+          <Textarea
+            value={alt}
             onChange={(_ev, data) => setAlt(data.value)}
             placeholder={t("(Optional)")}
+            style={{ height: "100px" }}
```
hadley, line 170: Could we also set `resize = {true}`?

mcanouil: Yes, manual resize is possible.

mcanouil: https://github.com/quarto-dev/quarto/pull/644/commits/615dc1ee4726a7be91400729cf55b7f10fcb4479

vezwork:
@mcanouil this looks good! Should we try to merge this? What would that take?

mcanouil:
I think this PR is quite safe as it does not change any behaviour.

I think it only takes an approval to merge this one that the UI change is good to go.

PR #649, opened 2025-02-03
fix: Ensure `#|` is added only at the beginning of a new line

### LINKED ISSUE #426: Auto addition of `#|` messes code on "return"
This can also happen in other less reproducible situations on save.
Currently, the way `#|` is not really reliable and most of the time causes issues rather than the opposite.

`#|` should only be added at the beginning of a new line.

https://github.com/quarto-dev/quarto/assets/8896044/0f4f67c6-b8a6-415b-b00e-a8c14d43da92

mcanouil:
Adjust the insertion logic to guarantee that `#|` is added at the start of a new line, addressing issues with code formatting on return. Fixes #426

Proposed behaviour:

https://github.com/user-attachments/assets/d3a58d6d-c2ce-4999-b028-b130f91b278e

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.119.0 (unreleased)
 
+- Ensure `#|` is added only at the beginning of a new line (<https://github.com/quarto-dev/quarto/pull/649>).
+
 ## 1.118.0 (Release on 2024-11-26)
 
 - Provide F1 help at cursor in Positron (<https://github.com/quarto-dev/quarto/pull/599>)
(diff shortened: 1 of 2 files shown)


mcanouil:
@cscheid @juliasilge gentle ping for this as it’s really a daily pain to have to fix the bad insertion.

mcanouil:
> I do think we'll eventually need to resolve the edit builder here with onEnterRules that other extensions can provide.

I do agree 100% with that but this means quite a change in the VSCode extension codebase which I was not willing to undertake at this point, especially without any tests.
The PR is really here to simply fix the current implementation not to improve/refactor it.

PR #655, opened 2025-02-08
fix: protect code cell options from formatting

### LINKED ISSUE #135: [vscode] Python formatter breaks comment pipe `#|` on code blocks
Currently, running a python formatter such as black on a qmd file destroys the comment pipe.

This code block

```py
#| echo: true
print("foo")
```

To

```py
# | echo: true
print("foo")
```

By auto formatting.

Perhaps the Quarto extension should protect the comment pipe as follows before letting the formatter run.

```py
# #| echo: true
print("foo")
```

### LINKED ISSUE #368: formatting document in VSCode with Quarto breaks option syntax
xref: https://github.com/quarto-dev/quarto-cli/issues/8576

### Bug description

I was told by @mcanouil here that I got my syntax wrong: https://github.com/quarto-dev/quarto-cli/discussions/8574#discussioncomment-8351297

I was wondering how that happened and figured it out. If you use the correct syntax and then format your document in VSCode using the Quarto extension, it adds a space. Since it's a single space in what looks like a Python comment, it's a bit hard to notice.

### Steps to reproduce

Create a cell in a Quarto document that has an option comment:

````
```{python}
#| panel: input
ui.input_text("species", "Species:", value="Gentoo")
```
````

use keybindings or right click to format document with -> choose Quarto:

image

image

click it, and a space appears!

image

### Expected behavior

I would not expect formatting the cell w/ the Quarto extension in VSCode to create syntax errors

### Actual behavior

formatting the cell w/ the Quarto extension in VSCode creates syntax errors

### Your environment

- IDE: VSCode w/ Quarto extension
- computer: Mac

### Quarto check output

```
Quarto 1.4.542
[✓] Checking versions of quarto binary dependencies...
      Pandoc version 3.1.11: OK
      Dart Sass version 1.69.5: OK
      Deno version 1.37.2: OK
[✓] Checking versions of quarto dependencies......OK
[✓] Checking Quarto installation......OK
      Version: 1.4.542
      Path: /Applications/quarto/bin

[✓] Checking tools....................OK
      TinyTeX: (not installed)
      Chromium: (not installed)

[✓] Checking LaTeX....................OK
      Tex:  (not detected)

[✓] Checking basic markdown render....OK

[✓] Checking Python 3 installation....OK
      Version: 3.11.7
      Path: /Users/cody/repos/pypi-analytics/venv/bin/python3
      Jupyter: 5.7.1
      Kernels: python3

[✓] Checking Jupyter engine render....OK

[✓] Checking R installation...........OK
      Version: 4.3.2
      Path: /Library/Frameworks/R.framework/Resources
      LibPaths:
        - /Library/Frameworks/R.framework/Versions/4.3-arm64/Resources/library
      knitr: 1.45
      rmarkdown: 2.25

[✓] Checking Knitr engine render......OK
```

mcanouil:
Implement a mechanism to prevent code cell options from being formatted, ensuring that only relevant code is affected during formatting operations.

Fixes #368, fixes #135

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -5,6 +5,7 @@
 - Added support for Positron's statement execution feature that reports the approximate line number of the parse error (<https://github.com/quarto-dev/quarto/pull/919>).
 - Fixed a bug where `Quarto: Format Cell` would notify you that no formatter was available for code cells that were already formatted (<https://github.com/quarto-dev/quarto/pull/933>).
 - No longer claim `.typ` files. Typst syntax highlighting in Quarto documents is unaffected, but standalone Typst files are now left to dedicated extensions like Tinymist (<https://github.com/quarto-dev/quarto/pull/943>).
+- Preserve Quarto code cell option directives (e.g. `#| label: foo`) when formatting embedded code. The directives are now stripped from the virtual document before being handed to the language formatter, so formatters such as Black, autopep8, and styler can no longer reflow or rewrite them (<https://github.com/quarto-dev/quarto/pull/655>).
 
 ## 1.130.0 (Release on 2026-02-18)
 
--- a/apps/vscode/src/providers/cell/options.ts
+++ b/apps/vscode/src/providers/cell/options.ts
@@ -91,7 +91,7 @@ function langCommentChars(lang: string): string[] {
     return chars;
   }
 }
-function optionCommentPattern(comment: string) {
+export function optionCommentPattern(comment: string) {
   return new RegExp("^" + escapeRegExp(comment) + "\\s*\\| ?");
 }
 
--- a/apps/vscode/src/providers/format.ts
+++ b/apps/vscode/src/providers/format.ts
@@ -37,15 +37,14 @@ import { isQuartoDoc } from "../core/doc";
 import { MarkdownEngine } from "../markdown/engine";
 import { EmbeddedLanguage, languageCanFormatDocument } from "../vdoc/languages";
 import {
+  isBlockOfLanguage,
   languageFromBlock,
   mainLanguage,
   unadjustedRange,
   VirtualDoc,
   virtualDocForCode,
-  virtualDocForLanguage,
   withVirtualDocUri,
 } from "../vdoc/vdoc";
-import { languageOptionComment } from "./option";
 
 export function activateCodeFormatting(engine: MarkdownEngine) {
   return [new FormatCellCommand(engine)];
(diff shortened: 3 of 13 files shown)


cscheid:
Another thought here is that this kind of PR _really_ makes me wish we had a test suite in this repo.

mcanouil:
> I assume this was done to prevent black and other Python auto-formatting tools from mangling our #| syntax. Is that the case? If so, I wonder if a more targeted, Python-only fix is better.

I was thinking about making it specific at first but then I realised that the expected comments are all hardcoded, so I reconsidered and make it a general protection for the comments symbols hardcoded (in two places).

> Another thought here is that this kind of PR _really_ makes me wish we had a test suite in this repo.

I'm not yet very familiar on this part, but I can't agree more with this.
That's on my personal roadmap before releasing Quarto Wizard 1.0.0.

- https://github.com/mcanouil/quarto-wizard/issues/5

mcanouil:
Let's hold on on this PR as the behaviour is not new at all so it can wait a bit for me to implement a test soon as I get some spare time.

mcanouil:
I've setup the test, but I can't get to make the formatting command to work.
I'm missing something, maybe related to vdoc or something else ...

juliasilge:
Ah, I don't think we can use any other extensions in the tests, because of the way that they use a special build of VS Code that is specially built for running extension tests.

Instead, what we'll want to do is mock out a formatter to run in the tests. This is the same thing that https://github.com/quarto-dev/quarto/pull/754 needs. We may want to put the formatting mock somewhere that all tests can use it.

mcanouil:
Well, `@vscode/test-cli` does support installing extensions and it is used here, even when calling via `yarn`.

I don't think the issue is the presence/absence of the formatter.

I can only make the test install the extension to be sure `yarn run test` has it when running.

mcanouil:
Ok, now I have some information. The vscode-test instance shows a notification that the selection marker is not within a cell that supports formatting which is clearly wrong, but no clue why `quarto` displays this.

mcanouil:
Even changing to your testFormatter function does not work.
Note that the formatting do work as it should but I really don't understand what is going on with this VSCode extension during testing.

See when running extension debug mode:

https://github.com/user-attachments/assets/6c9ffc36-a4d6-476e-9591-7b2603e4b4e9


I have to drop this as It costs me way too much time for no good reasons.

Feel free to take over as I won't spend more time on this for the next few months at least.

mcanouil:
I took another stab at this with some help this time (roborev + claude code for review).
It now works for select, code cell, and document-level formatting.

vezwork (approved):
Tested locally with the given test files and with the basic quarto [computation.qmd document from the tutorial website](https://quarto.org/docs/get-started/computations/positron.html). It works well.

The code looks quite safe and reasonable, the comments are very helpful. The tests also seem quite good. The body of the tests are a bit longer than I might prefer for human readability, but they seem like a good way of testing this. 

The tests and comments make me fairly confident that this works well and will continue to work well. Nice job @mcanouil!

PR #664, opened 2025-02-15
Add `allowIncomplete` to `runtime.executeCode()`

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/4459

Before this change, if you had some invalid code in a code chunk (R or Python) such as this:

````
```{r}
rnorm(1000, 1. 2)
```
````

then using "Run Cell" (either via the button or keybinding) in Positron would result in the console just looking like nothing is happening. What is happening under the hood is that Positron can tell the statement isn't "complete" and it's waiting for the user to finish it. That isn't an appropriate thing for the console to do in many situations, so we have the option `allowIncomplete`:

> Whether to bypass runtime code completeness checks. If true, the `code` will be executed by the runtime even if it is incomplete or invalid. Defaults to false.

This PR changes the Positron Quarto executors to use true, i.e. to skip the code completeness checks and just execute the code, even if it is about to error.


isabelizimm (approved):
lgtm! tried out the vsix and acts as expected in source and visual mode 🚀

Issue #675, opened 2025-03-07
Quarto vscode extension cell color match theme background


kylebutts:
I'm wondering if you all would welcome a small PR to the vscode extension to improve rendering of cells in quarto documents. I know that you can use `cells.background.light`/`cells.background.dark` to change the color, but this requires me to change this every time I switch themes. Adding an option `cells.background.useTheme` bool that would use `notebook.selectedCellBackground` theme color. This is a [built in color](https://code.visualstudio.com/api/references/theme-color#notebook-colors) for VSCode's notebooks and most themes will have a customized color. 

I think it would be a relatively easy change adding a branch here.
https://github.com/quarto-dev/quarto/blob/491057ba2005584b48ed511cf82f49d3b630362f/apps/vscode/src/providers/background.ts#L216-L234

See a similar patch to `positron-code-cells` extension: https://github.com/posit-dev/positron/pull/6633

kylebutts:
I assume this is good to close?

juliasilge:
Yes! Closed in #679

PR #676, opened 2025-03-07
Add option to use theme provided cell background (#675)

kylebutts:
Adds option `cells.background.useTheme` that when true will use the theme's built in cell background. 

Positron Light:
Positron Light
Gruvbox Light Soft:
Gruvbox Light Soft


juliasilge:
@kylebutts Are you able to undo the formatting changes in `package.json`?

kylebutts:
@juliasilge oops! Format on save did it!! See #675

PR #677, opened 2025-03-07
Add option to use theme provided cell background (#675)

kylebutts:
Adds option `cells.background.useTheme` that when true will use the theme's built in cell background. 

Positron Light:
Positron Light
Gruvbox Light Soft:
Gruvbox Light Soft

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -960,8 +960,16 @@
           "default": "#40404066",
           "markdownDescription": "CSS color for background of executable code cells on dark themes.\n\n*Note that this color should include an alpha channel so that selections show up against the background.*"
         },
-        "quarto.cells.background.delay": {
-          "order": 22,
+        "quarto.cells.background.useTheme": {
+					"order": 22,
+					"scope": "window",
+					"type": "string",
+					"format": "color",
+					"default": false,
+					"markdownDescription": "Should the background of executable code cells be based on the theme provided color.\n\nThis uses the `notebook.selectedCellBackground` color from the VSCode theme."
+				},
+				"quarto.cells.background.delay": {
+					"order": 23,
           "scope": "window",
           "type": "integer",
           "default": 250,
--- a/apps/vscode/src/providers/background.ts
+++ b/apps/vscode/src/providers/background.ts
@@ -213,8 +213,16 @@ class HiglightingConfig {
 
   public sync() {
     const config = vscode.workspace.getConfiguration("quarto");
-    const light = config.get("cells.background.light", "#E1E1E166");
-    const dark = config.get("cells.background.dark", "#40404066");
+    const useTheme = config.get("cells.background.useTheme", false);
+    let light, dark;
+    if (useTheme) {
+      const activeCellBackgroundThemeColor = new vscode.ThemeColor('notebook.selectedCellBackground');
+      light = activeCellBackgroundThemeColor;
+      dark = activeCellBackgroundThemeColor;
+    } else {
+      light = config.get("cells.background.light", "#E1E1E166");
+      dark = config.get("cells.background.dark", "#40404066");
+    }
 
     this.enabled_ = config.get("cells.background.enabled", true);
     this.delayMs_ = config.get("cells.background.delay", 250);


juliasilge:
The new `.vsix` is available here (download artifact, unzip, install VSIX):

https://github.com/quarto-dev/quarto/actions/runs/13730481250

apps/vscode/package.json
```diff
@@ -960,8 +960,16 @@
           "default": "#40404066",
           "markdownDescription": "CSS color for background of executable code cells on dark themes.\n\n*Note that this color should include an alpha channel so that selections show up against the background.*"
         },
-        "quarto.cells.background.delay": {
-          "order": 22,
+        "quarto.cells.background.useTheme": {
+					"order": 22,
+					"scope": "window",
+					"type": "string",
+					"format": "color",
```
juliasilge, line 967:
```suggestion
					"type": "boolean",
```

This is more like true vs. false, not a string formatted as a color, like what is show in this example:
https://code.visualstudio.com/api/references/contribution-points#Configuration-example

An example in this file is `"quarto.cells.background.enabled"`.

apps/vscode/package.json
```diff
@@ -960,8 +960,16 @@
...
+					"default": false,
+					"markdownDescription": "Should the background of executable code cells be based on the theme provided color.\n\nThis uses the `notebook.selectedCellBackground` color from the VSCode theme."
```
juliasilge, line 969:
```suggestion
					"markdownDescription": "Use the `notebook.selectedCellBackground` color from the current theme as the background of executable code cells."
```

juliasilge (commented):
Thank you so much for this PR @kylebutts! 🙌 

We are unfortunately getting to some dependencies between settings here, like the problem outlined at:
<https://design.tidyverse.org/independent-meaning.html>

The settings with the bad dependencies between them here are:

- `quarto.cells.background.enabled`
- the two colors `quarto.cells.background.light` and `quarto.cells.background.dark`
- the new `quarto.cells.background.useTheme`

If we were starting out from scratch, I would want to make the setting called `quarto.cells.background.enabled` an enum (find "enum / enumDescriptions / markdownEnumDescriptions / enumItemLabels" [here](https://code.visualstudio.com/api/references/contribution-points#contributes.configuration)) with options:

- "default" (maybe then change the colors to `quarto.cells.background.darkDefault` and `quarto.cells.background.lightDefault`)
- "useTheme"
- "off"

If we make that change now, it would be breaking for people who have set `quarto.cells.background.enabled` to false. I would imagine this is a very small number of people (no one has checked this setting in to [GitHub at least](https://github.com/search?q=quarto.cells.background.enabled&type=code)) and I personally would be willing to make such a breaking change.

@cscheid What do you think? I think this is a very good change in this PR, but where do you land on the relative badness of a breaking change in settings vs. confusing interaction of settings?

kylebutts:
Great, thanks for the feedback @juliasilge. I'll wait for feedback on your question and will either implement these changes or you suggested `quarto.cells.background.enabled` enumeration idea.

cscheid:
> @cscheid What do you think? I think this is a very good change in this PR, but where do you land on the relative badness of a breaking change in settings vs. confusing interaction of settings?

My personal philosophy wrt breaking changes.

- if we're going to break things, we should do it in bigger chunks. It's better to do bigger changes fewer times than smaller changes more times. The downside of breakage is that we inflict it on users, and so we should try to minimize the number of times they need to respond to our changes.
- The later we make breaking changes, the more I'm inclined to try and keep the previous functionality in. Fewer users => higher willingness to make the change

If I understand it right, not a lot of users appear to be exposed to the breakage here. In that case, I would go ahead and make a breaking change.

In this specific case, though, I think I would be confused by an `enabled` field that didn't support `true` or `false`.

juliasilge:
How about this set of settings?

- `quarto.cells.background`, an enum with options "default", "useTheme", and "off"
- `quarto.cells.background.darkDefault` and `quarto.cells.background.lightDefault`, which get ignored unless `quarto.cells.background` is "default"

juliasilge:
This is looking good, @kylebutts! 🙌 Thank you so much for iterating on this with us. 

It looks like something hasn't quite gone right and we have lost the defaults for `quarto.cells.background.delay` and the default background colors:

Screenshot 2025-03-10 at 3 54 55 PM

Can you take a look? It's possible it's the whitespace changes on 973 and 974? 🤷‍♀️ 

Also, can you make an update in `CHANGELOG.md`?

And yeah, it looks like 0e61dc247239c4375cce81c84e8845afb5a25c55 got applied as a new, separate commit with the same changes as e268ba45254913c8d2bfe2fe2bcca42577f21c32, so if you could revert that commit or start over or similar, that would be fantastic. Thank you again for your perseverance here!

kylebutts:
@juliasilge I think what's happening is that `quarto.cells.background` makes further `.` arguments (like `quarto.cells.background.delay`) not work. What do you think about `quarto.cells.background.display` as the option name for the choice of `default`/`useTheme`/`off`?

Issue #627, opened 2024-12-20
content deleted when switching between visual and text editor


isabelizimm:
[from bluesky](https://bsky.app/profile/joelnitta.com/post/3ldiegzhlhc2f)

> several paragraphs of code and prose went up in smoke when switching between visual and text editor... no commits made and nothing recoverable in the undo history

Unfortunately I have not been able to reproduce this, but can dig into it in the new year. I wonder if there's a way to swap editor types where it is not saved before switching.

joelnitta:
Thanks for opening this issue on my behalf. I didn't post it originally because it happened "in the moment" and I have not been able to replicate it. I was working primarily in the visual editor, then switched into source mode at least once. I lost the content when I switched into source mode. I had auto-save on and as mentioned in the OP the undo history didn't seem to work.

Positron Version: 2024.12.0 (Universal) build 96
Code - OSS Version: 1.93.0
Commit: c5ce275dc502f6b15433b271802cb33e1ba5ef68
Date: 2024-11-28T02:47:29.898Z
Electron: 30.4.0
Chromium: 124.0.6367.243
Node.js: 20.15.1
V8: 12.4.254.20-electron.0
OS: Darwin arm64 23.6.0

PR #684, opened 2025-03-20
Use new `positron.reopenWith` command to switch between visual/source editing

juliasilge:
Related to https://github.com/posit-dev/positron/pull/6698 where we added this new command, specifically to be used in Quarto

Also related to intermittent bad behavior we see reported such as in https://github.com/quarto-dev/quarto/issues/627

We've wanted to do this for a long time, because this new approach is much safer (more like the built-in `ReOpenInTextEditorAction` that VS Code itself uses) than trying to do all that switching from the extension. See more details in https://github.com/posit-dev/positron/issues/6230.

Unfortunately, we can't remove the other code path, because this command is one we newly exposed to extensions in Positron only.


juliasilge:
This also improves the behavior when you have _multiple_ editors open; with the other code path, you sometimes get weird closing of background editors.

juliasilge:
It's worth having someone try out the `.vsix` generated here in both VS Code and Positron; I see good behavior in both, although still some of the (same, expected) weirdness in VS Code unfortunately.

isabelizimm (approved):
the Positron experience is really nice! No more jittery editors 🎉

cscheid:
Long-awaited, this is really great!!

Issue #683, opened 2025-03-20
Quarto virtual documents intermittently overwrite contents of editor


juliasilge:
Originally posted by @brndngrhm in https://github.com/posit-dev/positron/discussions/6867

> Sometimes a quarto file in positron will get re-formatted in an unusable way. I just opened a project and a quarto doc i was working don no longer has any code chunks or YAML, and instead there many `#` inserted as new lines. I'm not sure what causes this and i can't seem to undo it, so i need to basically manually re-format the file by going through it, creating code chunks and then pasting code inside the code chunks

![image](https://github.com/user-attachments/assets/661b09c3-0199-412b-9f55-a64098f2d9d1)

There is an additional report by @kldhana in https://github.com/posit-dev/positron/discussions/6867#discussioncomment-12556576:

> I noticed this as well, but could not reproduce.

This looks like the Quarto virtual documents going haywire somehow, as pointed out by @DavisVaughan. In the GH discussion, it seems like it happens on saving with format-on-save turned on.

DavisVaughan:
@mvwestendorp were you also using Positron? Or just VS Code?

DavisVaughan:
@mvwestendorp if you have that small file with 1 code block where you saw this before and are willing to share it, it may help us reproduce

mvwestendorp:
> [@mvwestendorp](https://github.com/mvwestendorp) if you have that small file with 1 code block where you saw this before and are willing to share it, it may help us reproduce (even though we know Positron support isn't working right, this still seems like a very odd bug)

Sure it is not confidential, hopefully a gist is good way to share it. If not I am happy to hear how to share it in some other way.
 https://gist.github.com/mvwestendorp/a3245e30d48c12afdc326c72a8322e30

I do see the behavior again with that file. In the screenshots you see the results of repeatedly saving the file. It is still a bit inconsistent on whether it happens (it is quite odd indeed).

![Image](https://github.com/user-attachments/assets/e8c87d3a-971b-49a4-a3a1-4b3ebc6a19eb)

cscheid:
> Note I have completely uninstalled Air here

> I'm mostly convinced it is a Quarto timing bug now.

Yes, that seems right to me.

cscheid:
Wait, this _is_ the visual editor (_quarto_, not _quarto-cli_).

DavisVaughan:
OKAY Y'ALL - I did a deep dive into this, here's what I've found!

[...]

PR #688, opened 2025-04-02
Always use a unique virtual document file path

DavisVaughan:
Closes https://github.com/quarto-dev/quarto/issues/683

See full analysis in https://github.com/quarto-dev/quarto/issues/683#issuecomment-2773508072

I have managed to reliably (or at least, quickly) reproduce the problem by repeatedly mashing `Cmd + S` while also moving my mouse around over the code in this qmd (i.e. simultaneously invoking many hover requests). I have not been able to reproduce the issue this way with this PR.

``````

```{r}
#| eval: true
if (Sys.getenv("QUARTO_PROFILE") %in% "FR") {
  title <- "blablabla" # FR
  nomdesc <- "blablabla"
  MdcDesc <- "blablabla"
  patient_group_desc_h <- "blablabla"
  nomen_desc_h <- "blablabla"
  nomen_short_desc_h <- "blablabla"
} else {
  title <- "Executive summary" # NL
  nomdesc <- "blablabla"
  MdcDesc <- "blablabla"
  patient_group_desc_h <- "blablabla"
  nomen_desc_h <- "blablabla"
  nomen_short_desc_h <- "blablabla"
}
```

``````

### DIFF
--- a/apps/vscode/src/main.ts
+++ b/apps/vscode/src/main.ts
@@ -37,6 +37,10 @@ import { configuredQuartoPath } from "./core/quarto";
 import { activateDenoConfig } from "./providers/deno-config";
 
 export async function activate(context: vscode.ExtensionContext) {
+  // create output channel for extension logs and lsp client logs
+  const outputChannel = vscode.window.createOutputChannel("Quarto", { log: true });
+
+  outputChannel.info("Activating Quarto extension.");
 
   // create extension host
   const host = extensionHost();
@@ -84,7 +88,7 @@ export async function activate(context: vscode.ExtensionContext) {
     activateDenoConfig(context, engine);
 
     // lsp
-    const lspClient = await activateLsp(context, quartoContext, engine);
+    const lspClient = await activateLsp(context, quartoContext, engine, outputChannel);
 
     // provide visual editor
     const editorCommands = activateEditor(context, host, quartoContext, lspClient, engine);
@@ -125,6 +129,8 @@ export async function activate(context: vscode.ExtensionContext) {
 
   // activate providers common to browser/node
   activateCommon(context, host, engine, commands);
+
+  outputChannel.info("Activated Quarto extension.");
 }
 
 export async function deactivate() {
--- a/apps/vscode/src/vdoc/vdoc-tempfile.ts
+++ b/apps/vscode/src/vdoc/vdoc-tempfile.ts
@@ -29,45 +29,50 @@ import {
 } from "vscode";
 import { VirtualDoc, VirtualDocUri } from "./vdoc";
 
+/**
+ * Create an on disk temporary file containing the contents of the virtual document
+ *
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: apps/vscode/src/main.ts [hunk 1]
```diff
@@ -37,6 +37,10 @@ import { configuredQuartoPath } from "./core/quarto";
 import { activateDenoConfig } from "./providers/deno-config";
 
 export async function activate(context: vscode.ExtensionContext) {
+  // create output channel for extension logs and lsp client logs
+  const outputChannel = vscode.window.createOutputChannel("Quarto", { log: true });
+
+  outputChannel.info("Activating Quarto extension.");
```
DavisVaughan, line 43: I found it extremely useful to have an `outputChannel` to log to while debugging. I plumbed this through to points of interest so I could log.

I have since removed the log calls, but I still think this is quite useful for future us so I've left the actual output channel in to make this easier later on.

It is shared with the LSP client, essentially we provide our own output channel now rather than having the LSP client generate one automatically for us (we do the same thing in positron)

apps/vscode/src/vdoc/vdoc-tempfile.ts
DavisVaughan, line 1: I have streamlined this function quite a bit. I found that pretty much everything was shared code, and we don't actually need an "early exit" path for the `local` case - it's just the normal path tweaked to create the file in a known directory rather than a temp directory.

The diffs are a little large but I hope it is a big improvement overall

apps/vscode/src/vdoc/vdoc-tempfile.ts
```diff
@@ -82,19 +87,57 @@ async function deleteDocument(doc: TextDocument) {
...
+/**
+ * Generates a unique virtual document file path
+ *
+ * It is important for virtual documents to have unique file paths. If a static
+ * name like `.vdoc.{ext}` is used, it is possible for one language server
+ * request to overwrite the contents of the virtual document while another
+ * language server request is running (#683).
+ */
+function generateVirtualDocFilepath(directory: string, extension: string): string {
+  return path.join(directory, ".vdoc." + uuid.v4() + "." + extension);
```
DavisVaughan, line 142: Worth noting that for non-`local` LSP requests this was called `.intellisense.{uuid}.{ext}` but I do not think that name mattered. We now always use `.vdoc.{uuid}.{extension}`.

apps/vscode/src/vdoc/vdoc-tempfile.ts
DavisVaughan: Really the main actual change is that the `local` branch that used to exist used a static `.vdoc.{ext}` as the file path, which was being overwritten in the middle of an lsp request. It now uses `.vdoc.{uuid}.{ext}` to avoid this.

cscheid (approved):
Very good fix, LGTM.

The only comment I have is that this fix makes me wonder about other data races, and a code review might be in order...

PR #692, opened 2025-04-09
Display command titles on editor action bar in Positron

juliasilge:
Related to https://github.com/posit-dev/positron/pull/7191
Related to https://github.com/posit-dev/positron/issues/6440

I'll leave this in draft until we get some of the other controls set up as well. In VS Code, this all falls back to the basic behavior (no titles).

PR #698, opened 2025-04-16
Feature/Positron Editor Action Bar integration for Source / Visual toggle and [X] Render on Save checkbox

softwarenerd:
This PR updates the Quarto extension so that it places two new controls on the Positron Editor Action Bar. Those controls are:

- A toggle switch for changing between Source and Visual editing mode.
- A checkbox to enable or disable Render on Save for the current editing session.

Here's a video that shows these controls in action:

https://github.com/user-attachments/assets/af338522-c05e-491a-8591-17f89fbca2a4

The Source / Visual toggle control is quite straightforward.

The Render on Save checkbox control appears when a **Quarto** or **Quarto Shiny** document is open and it allows the user to independently override the configuration settings for **Quarto > Render: Render on Save** and **Quarto > Render: Render on Save Shiny**, depending on the document type, in the current editing session.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,5 +21,6 @@
     "eslint-config-custom": "*",
     "prettier": "^2.5.1",
     "turbo": "^1.8.5"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
 }
--- a/package.json
+++ b/package.json
@@ -21,6 +21,5 @@
     "eslint-config-custom": "*",
     "prettier": "^2.5.1",
     "turbo": "^1.8.5"
-  },
-  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
+  }
 }
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -426,6 +426,16 @@
         "enablement": "editorTextFocus && !editorReadonly && editorLangId == quarto"
       },
       {
+        "command": "quarto.toggleRenderOnSave",
+        "title": "Render on Save",
+        "category": "Quarto",
+        "enablement": "editorLangId == quarto",
+        "actionBarOptions": {
+          "controlType": "checkbox",
+          "checked": "(quarto.editor.type == quarto && quarto.editor.renderOnSave) || (quarto.editor.type == 'quarto-shiny' && quarto.editor.renderOnSaveShiny)"
+        }
+      },
+      {
         "command": "quarto.zoteroConfigureLibrary",
         "category": "Quarto",
         "title": "Zotero - Connect Web Library"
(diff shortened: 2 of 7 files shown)


package.json
```diff
@@ -21,5 +21,6 @@
     "eslint-config-custom": "*",
     "prettier": "^2.5.1",
     "turbo": "^1.8.5"
-  }
+  },
+  "packageManager": "yarn@1.22.22+sha512.a6b2f7906b721bba3d67d4aff083df04dad64c399707841b7acf00f6b133b7ac24255f2652fa22ae3534329dc6180534e98d17432037ff6fd140556e2bb3137e"
```
juliasilge, line 25: Is this change needed?

juliasilge (commented):
This is looking great! 🙌 Thank you so much for taking a look at how this plays out over in extension land.

With the change I made in 53a50e0de4525e57feb1ab2eed392139d0f1a4ea, this PR now closes #692. ✅ 

I want to highlight that there is some existing behavior that was already bad that now feels a bit worse, with the toggle between source and visual mode so prominently displayed now. If you have a document that specifies `editor: visual` in the YAML front matter and you are currently editing in visual mode, using any of our affordances to switch back to source editing _does not work_; you stay in visual mode. RStudio behaves in what I would consider the correct way; in this situation, if you switch the toggle, the editor switch to source. Do you think any of this new work could be used to fix that in this PR? Or should we open a followup issue to make that better?

juliasilge:
One other question for you @softwarenerd: If I open a Quarto file for the first time, the "Preview" button with its label shows up immediately, but the other two new controls take ~5 sec to show up. Is there anything we can do about that?

juliasilge:
I installed the `.vsix` from this PR into VS Code and everything looks as expected (i.e. no changes from here leaking through):

Screenshot 2025-04-16 at 7 36 47 PM

softwarenerd:
> One other question for you @softwarenerd: If I open a Quarto file for the first time, the "Preview" button with its label shows up immediately, but the other two new controls take ~5 sec to show up. Is there anything we can do about that?

I will continue to look into this. From everything I know at this point, this delay may be inherent in the startup of the extension. We draw these new components on the Editor Action Bar the moment they show up in React-land with a context key expression we can evaluate so we can determine how to draw them.

package.json
softwarenerd: I removed this. It's an artifact of the fact that I have yarn installed using `corepack enable`.

softwarenerd:
> This is looking great! 🙌 Thank you so much for taking a look at how this plays out over in extension land.
> 
> With the change I made in [53a50e0](https://github.com/quarto-dev/quarto/commit/53a50e0de4525e57feb1ab2eed392139d0f1a4ea), this PR now closes #692. ✅
> 
> I want to highlight that there is some existing behavior that was already bad that now feels a bit worse, with the toggle between source and visual mode so prominently displayed now. If you have a document that specifies `editor: visual` in the YAML front matter and you are currently editing in visual mode, using any of our affordances to switch back to source editing _does not work_; you stay in visual mode. RStudio behaves in what I would consider the correct way; in this situation, if you switch the toggle, the editor switches to source mode. Do you think any of this new work could be used to fix that in this PR? Or should we open a followup issue to make that better?

I will look into this. I wasn't aware of the `editor: visual` front matter setting (this was my first time messing with the Quarto extension), so I didn't test this scenario.

If the `editor: visual` front matter works the way the `render-on-save: true` front matter works, it supersedes everything. So this may be why none of our affordances for switching back to source editing work. I'll review and report back.

package.json
juliasilge: Hmmmm, this doesn't look removed quite yet on this PR. Can we get this change undone before we merge this PR?

juliasilge:
I opened https://github.com/posit-dev/positron/issues/7304 to explore further improvements in how these new UI affordances show up.

juliasilge:
I opened https://github.com/posit-dev/positron/issues/7305 for us to resolve the problem around switching to source mode if you are in visual mode.

PR #699, opened 2025-04-19
Improved (less aggressive) editor focus preservation behavior

jjallaire:
- Don't preserve focus during render on save as this often ends up stealing focus from the terminal (which can lead to stray characters in documents).
- Don't preserve focus after revealing preview window (we already tell vscode to preserveFocus -- this used to be inconsistently respected but now seems to work properly).

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.121.0 (unreleased)
 
 - Add new controls for Positron editor action bar (<https://github.com/quarto-dev/quarto/pull/698>).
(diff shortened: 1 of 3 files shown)


juliasilge (approved):
This is working well for me! It's a little bit of a bummer to not have the focus switch back to the editor, but I also have experienced that problem with stray characters and I agree that it is _much_ worse than having the user need to handle the focus themselves. Maybe we can revisit this in the future to get back some of that focus preservation.

I updated the changelog in 863233bda59b948fed3f2dd126f5385ba9322ab4 so this is good to merge now. ✅

PR #709, opened 2025-05-17
Add "insert code cell" to editor action bar

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/3923

This first go at this puts the new button on the right, includes the label, and uses the "insert" icon:

Screenshot 2025-05-17 at 5 19 56 PM

We can't fit it on the left, since adding it there pushes `"editor/actions/left"` to past the center mark, and the labels don't all fit and get borked. We may want to change how the labels on the editor action bar behave when they run out of space.

- Do we want to include the label? RStudio does not and seems like folks know how to use it.
- Do we want to use this "insert" icon? Another option would be the "plus" icon, or making a new one.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -296,8 +296,13 @@
       },
       {
         "command": "quarto.insertCodeCell",
+        "icon": "$(insert)",
         "title": "Insert Code Cell",
-        "category": "Quarto"
+        "category": "Quarto",
+        "actionBarOptions": {
+          "controlType": "button",
+          "displayTitle": true
+        }
       },
       {
         "command": "quarto.runSelection",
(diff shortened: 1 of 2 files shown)


apps/vscode/package.json
```diff
@@ -651,6 +656,13 @@
           "group": "1_editor_mode"
         }
       ],
+      "editor/actions/right": [
+        {
+          "command": "quarto.insertCodeCell",
+          "when": "editorLangId == quarto",
```
juliasilge, line 662: We don't need this in the Visual Editor (it has its own bar that includes inserting a code cell) or markdown files (executable code cells don't do anything in `.md`).

softwarenerd (approved):
LG. Sorry I spaced out on reviewing this. I was out sick on Monday and I lost track of the work item.

PR #710, opened 2025-05-18
Turn off visual mode completions in Positron

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/7201

This PR turns off the completions in the custom visual editor in Positron. I think the completions are similarly pretty bad in the visual editor in VS Code, but to be conservative, this PR only turns them off in Positron. If you look in that `codeview.ts` file, I think the completions are the only thing we really want to turn off for now, to avoid the truly heinous experience people are getting currently.

This does _not_ address https://github.com/posit-dev/positron/issues/7709 because the problem there is that the statement range provider isn't getting piped through to the visual editor at all; there is nothing to turn off there. The other not-so-great experiences we outline in https://github.com/posit-dev/positron/issues/1805 are because we are similarly _not_ providing something (for example, hover help), not because we are providing something badly that we can temporarily turn off.

## QA Notes

- You should get completions in the YAML headers of `.qmd`, in both visual and source
- You should still get completions in source editing, the same that you would get in a regular `.py` or `.R` file
- You should now _not_ get completions in a visual editor

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.122.0 (unreleased)
 
 - Change controls on Positron editor action bar and fix "Render on Save" behavior (<https://github.com/quarto-dev/quarto/pull/706>).
+- Turn off completions in visual mode in Positron, as a temporary stopgap until we can invest more in LSP features in the visual editor (<https://github.com/quarto-dev/quarto/pull/710>).
 
 ## 1.121.0 (Release on 2025-05-02)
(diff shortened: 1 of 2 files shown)


DavisVaughan (approved):
See https://github.com/quarto-dev/quarto/pull/712 first, which seems like a nice improvement in the clarity of this function. No behavior changes there.

Before this PR:

Screenshot 2025-05-19 at 12 49 52 PM

After this PR:

Screenshot 2025-05-19 at 12 50 01 PM

PR #712, opened 2025-05-19
Clarify that this is just a series of potential early exits

DavisVaughan:
Merges into #710 

No actual changes here, just reworking to make it easier to read as a series of early exiting if statements


cscheid:
heck yeah, team flat code reporting for duty http://iq0.com/notes/deep.nesting.html 🫡

Issue #708, opened 2025-05-13
Hundreds of `.vdoc`s appearing


DavisVaughan:
### Bug description

Reported in https://github.com/posit-dev/positron/issues/7664 by @ravichowdhury88

#### Positron and OS details:

Positron Version: 2025.06.0 (user setup) build 48
Code - OSS Version: 1.99.0
Commit: 621d1ce8f7fecc9623e9ab59b36c62f23d626d3d
Date: 2025-05-12T03:47:04.780Z
Electron: 34.3.2
Chromium: 132.0.6834.210
Node.js: 20.18.3
V8: 13.2.152.41-electron.0
OS: Windows_NT x64 10.0.26100

#### Interpreter details:

R-4.5.0

## Describe the issue:
Positron creating hundreds of .vdoc files in my code folder.

![Image](https://github.com/user-attachments/assets/38bf9d98-f14f-4f55-98e7-42a36d1b0c5f)

## Steps to reproduce the issue:

Not sure

## Were there any error messages in the UI, Output panel, or Developer Tools console?
NA

DavisVaughan:
This suggests that _someone_ is calling `virtualDocUriFromTempFile()` but is not calling the `cleanup` hook when its done with the uri
https://github.com/quarto-dev/quarto/blob/21d404d0ecb3ea28ca51b40e0e02c45e14dd7e96/apps/vscode/src/vdoc/vdoc-tempfile.ts#L74

We just have to figure out who it is...

DavisVaughan:
Because the files are visible, that implies they are `local` and are likely from `"format"` or `"definition"`

https://github.com/quarto-dev/quarto/blob/21d404d0ecb3ea28ca51b40e0e02c45e14dd7e96/apps/vscode/src/vdoc/vdoc.ts#L150

https://github.com/quarto-dev/quarto/blob/21d404d0ecb3ea28ca51b40e0e02c45e14dd7e96/apps/vscode/src/providers/format.ts#L181

https://github.com/quarto-dev/quarto/blob/21d404d0ecb3ea28ca51b40e0e02c45e14dd7e96/apps/vscode/src/lsp/client.ts#L283

The `"definition"` handler _looks_ fine. Like it has a `finally` block that does the cleanup.

One thing that caught my eye is that `"format"` uses `withVirtualDocUri()` and that does `vdocUri.cleanup()` without calling `await` on it like all other instances where we call `vdocUri.cleanup()`. I wonder if that has something to do with it.

[...]

PR #713, opened 2025-05-19
Remove `virtualDocUri()` as a footgun

DavisVaughan:
There were 3 remaining places where we used `virtualDocUri()` instead of `withVirtualDocUri()`. The ad hoc usage of this is scary because it requires us to carefully analyze the call to ensure we call the `cleanup()` hook (https://github.com/quarto-dev/quarto/issues/708). `withVirtualDocUri()` always handles this, so this PR removes `virtualDocUri()` as a footgun altogether by routing everything through `withVirtualDocUri()` and then unexporting `virtualDocUri()`.

### DIFF
--- a/apps/vscode/src/lsp/client.ts
+++ b/apps/vscode/src/lsp/client.ts
@@ -251,16 +251,13 @@ function embeddedSignatureHelpProvider(engine: MarkdownEngine) {
   ) => {
     const vdoc = await virtualDoc(document, position, engine);
     if (vdoc) {
-      const vdocUri = await virtualDocUri(vdoc, document.uri, "signature");
-      try {
-        return getSignatureHelpHover(vdocUri.uri, vdoc.language, position, context.triggerCharacter);
-      } catch (error) {
-        return undefined;
-      } finally {
-        if (vdocUri.cleanup) {
-          await vdocUri.cleanup();
+      return await withVirtualDocUri(vdoc, document.uri, "signature", async (uri: Uri) => {
+        try {
+          return await getSignatureHelpHover(uri, vdoc.language, position, context.triggerCharacter);
+        } catch (error) {
+          return undefined;
         }
-      }
+      });
     } else {
       return await next(document, position, context, token);
     }
--- a/apps/vscode/src/lsp/client.ts
+++ b/apps/vscode/src/lsp/client.ts
@@ -23,6 +23,7 @@ import {
   LocationLink,
   Definition,
   LogOutputChannel,
+  Uri
 } from "vscode";
 import {
   LanguageClient,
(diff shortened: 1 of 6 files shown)


apps/vscode/src/lsp/client.ts
```diff
@@ -225,19 +226,13 @@ function embeddedHoverProvider(engine: MarkdownEngine) {
...
-      try {
-        return getHover(vdocUri.uri, vdoc.language, position);
-      } catch (error) {
-        console.log(error);
-      } finally {
-        if (vdocUri.cleanup) {
-          await vdocUri.cleanup();
+      return await withVirtualDocUri(vdoc, document.uri, "hover", async (uri: Uri) => {
+        try {
+          return await getHover(uri, vdoc.language, position);
```
DavisVaughan, line 231: We did not `await` before but I'm guessing it's probably better to?

apps/vscode/src/lsp/client.ts
```diff
@@ -255,16 +250,13 @@ function embeddedSignatureHelpProvider(engine: MarkdownEngine) {
...
-      try {
-        return getSignatureHelpHover(vdocUri.uri, vdoc.language, position, context.triggerCharacter);
-      } catch (error) {
-        return undefined;
-      } finally {
-        if (vdocUri.cleanup) {
-          await vdocUri.cleanup();
+      return await withVirtualDocUri(vdoc, document.uri, "signature", async (uri: Uri) => {
+        try {
+          return await getSignatureHelpHover(uri, vdoc.language, position, context.triggerCharacter);
```
DavisVaughan, line 255: We did not `await` before but I'm guessing it's probably better to?

apps/vscode/src/lsp/client.ts
```diff
@@ -280,64 +272,61 @@ function embeddedGoToDefinitionProvider(engine: MarkdownEngine) {
   ): Promise<Definition | LocationLink[] | null | undefined> => {
     const vdoc = await virtualDoc(document, position, engine);
     if (vdoc) {
-      const vdocUri = await virtualDocUri(vdoc, document.uri, "definition");
```
DavisVaughan, line 283: This diff looks ugly but I didn't change anything in the `try` or `catch` block. It's just switching to `withVirtualDocUri()` and removing the `finally` block

apps/vscode/src/lsp/client.ts
juliasilge: I _think_ you are correct on this; I occasionally would observe slight glitchy-ness with the embedded help in Quarto files and I can't get it to happen now with this PR.

juliasilge (approved):
Thank you for working on this! Super happy to see those `virtualDocUri` uses cleaned up 🙌

Issue #689, opened 2025-04-03
Edit in Visual Mode broken when using Quarto extension 1.119.0 with Positron 2025.03.0-116


sharon-wang:
### Bug description

Originally opened by @memeplex in https://github.com/posit-dev/positron/issues/7098.

In `Positron 2025.03.0-116` with `quarto extension 1.119.0`, when editing a qmd file, nothing happens when <kbd>Edit in Visual Mode</kbd> is used from the shortcut, menu or command.

I can reproduce this in Positron 2025.03.0-116 as well, but the issue no longer occurs in Positron [2025.04.0-173](https://github.com/posit-dev/positron/releases/tag/2025.04.0-173).

I think what happened is that the Quarto extension auto-updated to 1.119.0 which uses `positron.reopenWith`, however `positron.reopenWith` was not yet implemented in `2025.03.0-116` (hence the command not found).

### Steps to reproduce

1. Create a qmd file
2. Try to open visual editor

### Actual behavior

File does not open in visual editor and error occurs

The Extension Host output shows this:

```
[error] Error: command 'positron.reopenWith' not found
    at $b5b2.n (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:340565:29)
    at $b5b2.executeCommand (vscode-file://vscode-app/Applications/Positron.app/Contents/Resources/app/out/vs/workbench/workbench.desktop.main.js:340560:17)
```

### Expected behavior

The visual editor is opened.

---

If the `executeCommand` fails, we should fallback to the previous save and open editor behaviour.

https://github.com/quarto-dev/quarto/blob/a410a3c2bd78a1e1398c033d6c8f96a871617f02/apps/vscode/src/providers/editor/toggle.ts#L104-L110

### Your environment

macOS Sequoia 15.1

#### Positron and OS details:

Positron Version: 2025.03.0 (Universal) build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:51:44.014Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Darwin arm64 24.1.0

juliasilge:
[...]

PR #736, opened 2025-06-17
Specify a required version of Positron

juliasilge:
Related to https://github.com/posit-dev/positron/pull/8058

I was a bit on the fence about what to require here and am very open to feedback. 

- The absolute minimum would be 2025.04.0, to protect against the failure seen in https://github.com/quarto-dev/quarto/issues/689. 
- I would love to get a version that includes https://github.com/posit-dev/positron/pull/7860, but that first showed up in build 2025.06.0-167 so I think we'd have to do 2025.07.0 (right?) and that is too aggressive as of today.

Thoughts?

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -28,7 +28,8 @@
   ],
   "private": true,
   "engines": {
-    "vscode": "^1.75.0"
+    "vscode": "^1.75.0",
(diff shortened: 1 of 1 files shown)


softwarenerd (approved):
LG!

PR #740, opened 2025-06-19
Add pandoc path lookup and optional setting

fflores97:


cderv:
Hi! Thanks for your contribution. 

Can you reexplain the problem you are trying solve? I understand this is because of Nix specific build and setup for QUarto, right? 

To remind you of the context, Quarto works with an expected version of Pandoc. Quarto's installer does bundle Pandoc, and the VSCODE extension currently looks in there. 
Nix has a specific setup because it rebuilds Quarto by using an external Pandoc. 
Quarto allows that through `QUARTO_PANDOC` environment variable, but quarto is still meant to work with the expected version of Pandoc. When use with another version, there is no warranty of 💯 % working. 

Nix package does set `QUARTO_PANDOC`: https://github.com/NixOS/nixpkgs/blob/9cd27c52f82d1d8122b11c5c3556cccc12805d86/pkgs/development/libraries/quarto/default.nix#L62

So I wonder if the "fix" should be to look for `QUARTO_PANDOC` env var if Pandoc is not found in the expected place inside the Quarto install folder. 

Nix could also add symlink in the expected place.    

I hope it makes sense. I am trying to prevent side effects if the VSCode extension starts to detect unwanted Pandoc.

fflores97:
Thank you, Christophe. The problem is that the NixOS quarto package doesn't bundle pandoc inside it like in other distributions. In fact, nixpkgs [explicitely deletes it](https://github.com/NixOS/nixpkgs/blob/9cd27c52f82d1d8122b11c5c3556cccc12805d86/pkgs/development/libraries/quarto/default.nix#L77), so the [default](https://github.com/quarto-dev/quarto/blob/2c497155a2f2f62d87e2531e2fb996ae6aa3e20c/packages/quarto-core/src/context.ts#L83) `pandocPath` variable won't consider other pandoc paths.

I like your idea better, though. As far as I can tell, currently nothing in the codebase uses `QUARTO_PANDOC`, but I see it as a good chance to use it. I just tried this and could get visual edit mode to work with just the env variable. Very elegant!

cderv:
Yes, I know they remove it - they set the env var to the real location. They could also have done symlink I guess. 

The env var is the expected advanced way to change Pandoc used by Quarto, so I would use this instead of trying to detect outside of Quarto.

vezwork:
I pulled this PR and tested it locally. I built with `yarn run dev-vscode` and ran an extension host. Previewing and using the visual editor continue to work, it does not seem to cause any issues.

Looks good to me.

vezwork (approved):

cderv:
Thanks for the clarification @cscheid ! 

Still ok to merge this right ?

cscheid:
@fflores97 Thanks for the contribution, we will merge it. Although we don't need a contributor agreement for a small PR like this, you might want to send one so that's one fewer step in the future. If you're willing to do that, we have instructions you can follow [in our website](https://quarto.org/about.html#contribute).

vezwork:
Note for the future: I think it would have made the git history nicer if we rebased on main before merging

cderv:
ok noted ! thanks for the feedback

juliasilge:
@fflores97, we are trying to track you down to chime in on https://github.com/quarto-dev/quarto/issues/1026. 

We are systematically re-licensing Quarto editor components to use the MIT license, to make our licenses as clear and permissive as possible. To do so, we need the approval of all copyright holders, which I have found by reviewing contributions from all all non-Posit contributors. 

Would you permit us to re-license the Quarto VS Code extension and related components in this repo with the (more permissive) MIT license? If so, take a look at the linked issue ⤴️ and please comment "I agree".

PR #748, opened 2025-07-02
Format vscode extention and LSP files consistently

lionel-:
The TS files in the LSP and VS Code extension are currently inconsistently indented, sometimes with spaces, sometimes with tabs. As I work in Zed, an editor that doesn't have automatic detection of indentation, it's very inconvenient for me to write code in these files.

So I've reformatted all files using VS Code's TS formatter (which I can also use from Zed) thanks to this great extension: https://marketplace.visualstudio.com/items?itemName=lacroixdavid1.vscode-format-context-menu. The internal formatter is limited but I think this is an advantage as it's mostly inconsistent whitespace that has been changed in this PR. Reformatting with Biome or Prettier would have far more impact and so likely require a little more thought and discussion.

I know it's always a bummer and a potential source of conflicts to resort to a large commit that sprinkles the repo of trivia changes, but I think this will significantly improve workflow as we contribute to the Quarto extension.

Another change in this PR is that Zed gains a project settings file, and I've disabled automatic detection of indentation in VS Code settings so that it always use the configured indentation style and we don't fall back into the inconsistent state. Both IDEs are now configured to use spaces. I've checked that this was the main indentation style in `apps/lsp` and `apps/vscode`.


lionel-:
@cscheid Most people I know are using VS Code's language server for TS development, and it's possible to use it in other editors. We do have it enabled by default, see: https://github.com/quarto-dev/quarto/blob/bea15fdea220e4545477be4882da117679485dd4/.vscode/settings.json#L4

The main issue fixed here is the inconsistent indentation style. A potential cause of this is the combination of having copied over files from other projects, and VS Code respecting the document's indentation instead of the project setting. This is what this PR fixes.

lionel-:
Gotcha I'll take a look at Deno and make sure that it's configured to use 2 spaces like the `apps/` projects formatted in this PR.

I don't mind switching to Deno for formatting these sub projects but that could be the subject of an ulterior PR. This one is mostly whitespace.

By the way I learned from @DavisVaughan today that we can add a file like this to exclude style-change commits from blame (we need the commit hash so that would have to be added later): https://github.com/biomejs/biome/blob/main/.git-blame-ignore-revs

cscheid:
I think we're good to merge this.

I reviewed all the files in the PR and in the monorepo as well, and I think we've succesfully contained the Deno LSP to quarto-cli (and that's a good thing. We shouldn't let that escape containment.)

cscheid (approved):

cscheid:
I'm sorry to bring this up again, but I don't think this PR solves the problem. For example, I still see on `main`, `package.json`, under the script that would run for `npm run format`:

```
    "format": "prettier --write \"**/*.{ts,tsx}\"",
```

Then, we have .zed/settings.json that needs to be maintained in conjunction with .vscode/settings.json, and both use something else besides prettier.

I would prefer that we use external (ie, independent of any particular editor) tooling consistently for this repo, and that for typescript code, we all decided on a _single_ top-level .prettierrc file that is then just used everywhere consistently. VS Code supports directly formatting through prettier via an extension (https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode), but I don't know about Zed etc.

lionel-:
@cscheid Sounds good to me! I can take a look at setting this up in IDEs and updating the LSP and extension files, once the current PRs are merged. I'd prefer let someone else reformat the other parts of the project though.

cscheid:
> I'd prefer let someone else reformat the other parts of the project though.

Yeah, agreed.

My general preference for formatting changes is to delay them until there are additional non-formatting changes to be committed. With that said, the _reason_ I have this preference is that pure-formatting commits do not play nicely with `git blame`, and I use `git blame` daily. However, @vezwork found [this technique](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/#ignore-commits-via-%60--ignore-revs-file%60) to teach `git blame` to skip specific commits.

So if we have a global `.git-blame-ignore-revs` file and a corresponding `git config blame.ignoreRevsFile` entry in the repo, then I would be much more receptive to trivia-only commits. They would still pollute the summary entry on the GitHub UI directory listing, but not showing up on `git blame` is huge.

lionel-:
I understand the hesitations around format commits, but in the age of format-on-save there doesn't seem to be a way around it unfortunately

> My general preference for formatting changes is to delay them until there are additional non-formatting changes to be committed. 

This works well when the user is restyling, but not as much when it's a formatter set for the whole file. I think it's more distracting to have lots of unrelated style changes in a commit that extend far beyond the substantive changes, compared to a single format commit that you can ignore as a whole.

> owever, @vezwork found [this technique](https://www.stefanjudis.com/today-i-learned/how-to-exclude-commits-from-git-blame/#ignore-commits-via-%60--ignore-revs-file%60) to teach git blame to skip specific commits. So if we have a global .git-blame-ignore-revs file and a corresponding git config blame.ignoreRevsFile entry in the repo, then I would be much more receptive to trivia-only commits.

@DavisVaughan also let me know about this, and I added one already!
https://github.com/quarto-dev/quarto/blob/main/.git-blame-ignore-revs

DavisVaughan:
> My general preference for formatting changes is to delay them until there are additional non-formatting changes to be committed. With that said, the reason I have this preference is that pure-formatting commits do not play nicely with git blame, and I use git blame daily.

@cscheid in the most respectful way possible, I find this so crazy 😆. I feel like the optimal way to handle adopting a new formatter is:
- 1 PR that changes everything that gets squashed into 1 commit
- `.git-blame-ignore-revs` ignores that 1 commit

Then you're done, and you never have to worry about having a too-large diff when you come back and touch a file that hasn't been changed in years and hasn't been updated to the new formatting style yet. As a PR reviewer, having to look at a diff like that involving both lots of formatting changes + the actual behavior change is incredibly frustrating. Better to get all formatting changes out of the way in one go, and never think about it ever again.

PR #749, opened 2025-07-03
Build LSP directly in destination so source maps are correct

lionel-:
We were building inside `apps/lsp/dist`, then copying over to `apps/vscode/out/lsp`. But the source maps are relative so they were no longer pointing to the correct locations.

Since this is a mono repo, I thought we could just directly build in the vscode's output directory.

With this change, we can set breakpoints inside the LSP. You'll first need to use the "Attach to LSP" launch action once the LSP process has been spawned.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -6,11 +6,13 @@ on:
     branches:
       - main
     paths:
+      - 'apps/lsp/**'
       - 'apps/vscode/**'
   pull_request:
     branches:
       - main
     paths:
+      - 'apps/lsp/**'
       - 'apps/vscode/**'
   workflow_dispatch:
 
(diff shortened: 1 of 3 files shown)


lionel-:
Good point, I've checked that `vsce package` still works locally but I should do a manual run of the action on this PR.

cscheid:
Do similar changes need to happen in https://github.com/quarto-dev/quarto/blob/main/.github/workflows/publish.yaml? There's references to apps/vscode there as well.

juliasilge (approved):
It took me a couple of tries to get it to work, but I did manage to set a breakpoint in the LSP with this PR.

How would you feel about adding something in `README.md` or a new `CONTRIBUTING.md` about how to get that set up?

I cannot tell if we are building `apps/vscode` twice. 😬

PR #750, opened 2025-07-03
Fix logging and configuration watching in LSP

lionel-:
The LSP is currently not emitting any log messages because:

- The default log level is 'off'
- We're not watching any log level setting so it can't be changed

I've also noticed the following issues:

- The log level setting on the LSP side is `markdown.server.log` which is a remnant from copying over the Markdown LSP.

- The LSP log levels are currently "off", "debug", and "trace". I think we're missing "warn" and "error" for important user messages, and "info" for benign user messages. If we have these, then we should probably remove "off" and have the least verbose level set to "error" (this is what we do in Ark for instance).

Main fixes in this PR:

- Move log level setting to `quarto.server.logLevel`. Since it's now a `quarto` setting, we get notified when it changes.
- Handle `logLevel` change events on the server side so the LSP can immediately react to log level changes.
- Also send `logLevel` in `initializationOptions` so the LSP can use the level set by the user as early as possible.

Adjacent fixes and cleanups:

- Refactor log levels to take error, warn, info, debug, and trace levels.
  Set default trace level to `warn`.
- Add `log()` wrappers like `logError()`, `logWarn()`, etc, as well as `logNotification()` and `logRequest()` to log at trace level received LSP messages.
- Call the latter in message handlers to increase the coverage of LSP events at trace level, which is very valuable for debugging.

- Minor: The logger class currently emits messages via `console.log`, which `languageserver` redirects to the client when the connection is of type StdIo (see // https://github.com/microsoft/vscode-languageserver-node/blob/df56e720/server/src/node/main.ts#L262-L264). I changed this to explicitly send messages via our connection object instead, which should also work if someone needs to connect to the LSP via TCP. This will also allow us to use more accurate LSP loggers once we switch to languageclient 10 which will support `LogChannelOutput` on the client side (see https://github.com/microsoft/vscode-languageserver-node/issues/1116).

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -1,0 +1,8 @@
+# This file lists revisions of large-scale formatting/style changes so that
+# they can be excluded from git blame results.
+#
+# To set this file as the default ignore file for git blame, run:
+#   $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Format vscode extention and LSP files consistently (#748)
+a1513effc913e6e59651256d79295da37134dbbf
--- a/apps/lsp/.eslintrc.js
+++ b/apps/lsp/.eslintrc.js
@@ -1,4 +1,7 @@
 module.exports = {
   root: true,
   extends: ["custom-server"],
+  rules: {
+    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
+  },
 };
--- a/apps/vscode/justfile
+++ b/apps/vscode/justfile
@@ -1,0 +1,8 @@
+install:
+  rm -rf *.vsix && vsce package && code --install-extension *.vsix && positron --install-extension *.vsix
+
+install-vscode:
+  rm -rf *.vsix && vsce package && code --install-extension *.vsix
+
+install-positron:
+  rm -rf *.vsix && vsce package && positron --install-extension *.vsix
--- a/apps/vscode/justfile
+++ b/apps/vscode/justfile
@@ -1,8 +1,0 @@
-install:
-  rm -rf *.vsix && vsce package && code --install-extension *.vsix && positron --install-extension *.vsix
-
-install-vscode:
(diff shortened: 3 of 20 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# This file lists revisions of large-scale formatting/style changes so that
+# they can be excluded from git blame results.
+#
+# To set this file as the default ignore file for git blame, run:
+#   $ git config blame.ignoreRevsFile .git-blame-ignore-revs
+
+# Format vscode extention and LSP files consistently (#748)
+a1513effc913e6e59651256d79295da37134dbbf
```
lionel-, line 8: This should skip the format commit in blame, at least on github

apps/lsp/.eslintrc.js
```diff
@@ -1,4 +1,7 @@
 module.exports = {
   root: true,
   extends: ["custom-server"],
+  rules: {
+    '@typescript-eslint/no-unused-vars': ['error', { argsIgnorePattern: '^_' }],
```
lionel-, line 5: Don't lint unused variables prefixed with `_`

apps/vscode/justfile
```diff
@@ -0,0 +1,8 @@
+install:
```
lionel-, line 1: Taking example on Biome, we've started using the `just` command runner in Ark and Air. I've started one here to easily update dev Quarto in Code and Positron.

juliasilge (commented):
What I am seeing when I build from this PR is a bit confusing:

Screenshot 2025-07-04 at 2 46 29 PM

I see Quarto LSP logging coming through 🎉 but the default for this setting seems to be warning while I see even trace messages coming through. What do you think might be happening here?

Also, like I said in #753 I think the `just` file should be at the root with the other build and dev scripts.

lionel-:
@juliasilge Thanks for catching that, this should now work as expected with the user log level taking effect immediately on init. I've removed the justfile (it was predating the contributing PR).

PR #751, opened 2025-07-03
Exclude section headers from workspace symbols in R packages

lionel-:
Addresses https://github.com/posit-dev/positron/issues/8401

Once activated, the Quarto extension currently scans all md files and exports section headers to workspace symbols. While great generally, this can get in the way of the dev workflow in R packages.

R packages may include a variety of md files:

- Snapshot files: https://github.com/tidyverse/ellmer/tree/main/tests/testthat/_snaps
- Documentation: https://github.com/r-lib/rlang/tree/main/man/rmd

This causes Quarto to flood workspace symbols with headers that are mostly irrelevant for package development:

https://github.com/user-attachments/assets/a0269da3-68b5-44b2-b10c-ccae9a9196b3

I think by default (without any user configuration), headers shouldn't be exported as workspace symbols in R packages. But we need to preserve the existing behaviour for other projects.

To fix this, this PR:

- Introduces a new setting `quarto.symbols.exportToWorkspace`, with values `"default"`, `"all"`, or `"none"`. We need an enum with a default case instead of a boolean to provide the default behaviour that depends on the project type. The other two values allow hardcoding the behaviour in a project.

- When set to `"default"`, the LSP detects if the project is an R package.

  While the LSP would ideally be agnostic regarding the project type, having a customisation mechanism for extensions to set the default behaviour for a project would be complicated and require conflict solving if two extensions try to handle the same project.

- We already have a tool to detect R packages in the extension module (thanks to @juliasilge to have let me know!). It's a bit complicated because we need to parse the project's DESCRIPTION file if present, so I think we need to use the same implementation in both the LSP and extension. I pulled it up one level in `apps/utils`. To keep things light I didn't make it a full node module.

- The tool in `apps/utils` take a folder path. Then two wrappers in `apps/lsp` and `apps/vscode` are in charge of unpacking the relevant workspace folder (via different APIs specific to the LSP and extension) and calling the util.

### QA Notes

Create an R package with markdown files and make sure the Quarto extension is activated (it activates when navigating to an R file). Look for workspace symbols (cmd T, or `#` prefix in command palette). The markdown headers shouldn't be included.

In a project that's not an R package, they should be included.

In the R package you can opt into including the headers via `quarto.symbols.exportToWorkspace`. In the regular project you can disable them. The setting should immediately apply (at the next request).

### DIFF
--- a/apps/lsp/src/config.ts
+++ b/apps/lsp/src/config.ts
@@ -43,6 +43,9 @@ export interface Settings {
       readonly scale: number;
       readonly extensions: MathjaxSupportedExtension[];
     }
+    readonly symbols: {
+      readonly exportToWorkspace: 'default' | 'all' | 'none';
+    };
   };
   readonly markdown: {
     readonly preferredMdPathExtensionStyle: 'auto' | 'includeExtension' | 'removeExtension';
@@ -88,6 +91,9 @@ function defaultSettings(): Settings {
       mathjax: {
         scale: 1,
         extensions: []
+      },
+      symbols: {
+        exportToWorkspace: 'all'
       }
     },
     markdown: {
@@ -169,6 +175,9 @@ export class ConfigurationManager extends Disposable {
         mathjax: {
           scale: settings.quarto.mathjax.scale,
           extensions: settings.quarto.mathjax.extensions
+        },
+        symbols: {
+          exportToWorkspace: settings.quarto.symbols.exportToWorkspace
         }
       }
     };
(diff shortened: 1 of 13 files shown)


lionel-:
oops, it seems this got auto-merged by github _into the LSP log branch_ after I messed up a rebase and included these commits into the other branch by mistake. Now fixed on the log branch side, I'll have to create a new PR as it doesn't seem I can reopen this one. PRs for stacked branches are hard 😬 

Screenshot 2025-07-07 at 11 26 23

🤔 ??

PR #753, opened 2025-07-04
Add `CONTRIBUTING.md` in vscode extension and top-level install commands

lionel-:
Taking example on Biome, we've started using the just command runner in Ark and Air. I've started one in the vscode extesion to easily install it in Code and Positron and build it in dev/watch mode.

The new contributing file describes our dev workflows.


juliasilge (commented):
All the other runner type support in this repo is at the _root_. Can we set up the `just` file so that it also runs from the root, for consistency?

What do you think about putting `CONTRIBUTING.md` in [its regular location in `.github` or also at the root](https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors#adding-a-contributing-file)? I wouldn't know to look for it where it currently is.

lionel-:
On the other hand, having a justfile at the top-level is a different matter since, as you mention, we already have a top-level workflow. Are we ready to add a justfile at top-level in addition to our existing `yarn` commands? To avoid any confusion, I've removed the justfile and added the relevant scripts as top-level `yarn` commands.

The top-level readme now links into our project's contributing file for easier access.

Issue #745, opened 2025-06-30
Reformating R cell in VSCODE with air does not correctly close chunk


cderv:
Originally posted at 
- https://github.com/posit-dev/air/issues/374#issuecomment-3016002896

But it seems Quarto: Format Cell is a feature from Quarto extension. 

----

Not sure where to post this finding, if this is Quarto VSCODE extension that integrates with Air, or Air VSCODE extension that does.

Here is the document
````markdown
---
title: "kableExtra"
format: html
execute: 
  eval: false
---

## Quarto

```{r}
library(quarto)

# Without helper functions - Markdown won't be processed
basic_data <- data.frame(
  Item = c("Item 1", "Item 2", "Item 3"),
  Description = c("**Bold text**", "*Italic text*", "`Code text`")
)

knitr::kable(basic_data, format = "html", escape = FALSE, caption = "Without Quarto processing")
```
````

And here is what happens

https://github.com/user-attachments/assets/0e210a91-3d25-4bb2-8251-2535042d7c51

The closing backticks are not correctly placed, and the last parenthesis is removed

juliasilge:
@math-mcshane Can you see if you still observe the same problems when there are newlines around your code chunks? For example, add an extra blank line between lines 163 and 164 (and similar throughout). We are noticing that folks may not be aware of the fact that pandoc markdown requires these kinds of empty lines, as specified here:

<https://pandoc.org/MANUAL.html>

Note:

> Like regular code blocks, fenced code blocks must be separated from surrounding text by blank lines.

We have https://github.com/quarto-dev/quarto-cli/issues/12255 logged to improve the documentation around this and plan to improve our parsing of markdown.

math-mcshane:
@juliasilge do you mean that pandoc requires the new lines to render? I'm able to render this document to reveal.js HTML and web page ([visible here](https://stat-37815-au24.ryanmcshane.com/l03-baserfundamentals#console-output)) without issue. If pandoc needs this, is knitr or something else adding the spaces to make it render? 

(AFK, but will check whether Air still tries to split the closing backticks when I get home.)

math-mcshane:
[...]

PR #754, opened 2025-07-04
Fix cell formatting issues

lionel-:
Addresses https://github.com/quarto-dev/quarto/issues/745

But needs to be tested more widely and deal with formatters that don't ensure a trailing newline as discussed in https://github.com/quarto-dev/quarto/issues/745#issuecomment-3035444418.


apps/vscode/src/providers/format.ts
```diff
@@ -138,9 +138,13 @@ class FormatCellCommand implements Command {
       const edits = await formatActiveCell(editor, this.engine_);
       if (edits) {
         editor.edit((editBuilder) => {
-          edits.forEach((edit) => {
-            editBuilder.replace(edit.range, edit.newText);
-          });
+          // Sort edits by descending start position to avoid range shifting issues
```
lionel-, line 141: Unrelated but I thought we should make this robust to providers returning unsorted edits.

apps/vscode/src/providers/format.ts
```diff
@@ -224,8 +233,17 @@ async function formatBlock(doc: TextDocument, block: TokenMath | TokenCodeBlock,
           new Position(edit.range.end.line + block.range.start.line + 1, edit.range.end.character)
         );
         return new TextEdit(range, edit.newText);
-      })
-      .filter(edit => blockRange.contains(edit.range));
+      });
+
+    // Bail if any edit is out of range. We used to filter these edits out but
+    // this could bork the cell.
+    if (edits.some(edit => !blockRange.contains(edit.range))) {
```
lionel-, line 240: Unrelated but I'm pretty sure applying edits partially can bork our cells 😬

cscheid:
Thanks for the PR! Something for us to consider: 

Every time I see a PR here I get a feeling of dread: we have no test coverage and so I have very little visibility or intuition of the impact that these changes make. What kind of infrastructure would we have to add for us to have unit test coverage for these kinds of changes?

lionel-:
Part of the reason this is still a draft is that I'd like to add some tests here. They'll be limited in scope though.

Here are the levels of tests we have in our R stack, by ascending level of scope:

- Internal LSP tests, these are easiest to write.
- Protocol-level tests for LSP with a custom client.
- Extension tests. With these you can test effects of vscode commands on e.g. text in an editor. These commands may delegate to the LSP. Example: https://github.com/posit-dev/positron/blob/main/extensions/positron-r/src/test/indentation.test.ts
- UI tests in the Positron suite: https://github.com/posit-dev/positron/tree/main/test/e2e/tests/quarto

We spend most of our time and energy on internal LSP tests.

vezwork:
I pulled down this PR and followed the reproduction steps in #745 (which this PR should address). I ran into a new error:

Screenshot 2025-09-04 at 12 12 30 PM

I was able to reproduce #745 in Positron on the `main` branch of the extension, so this error is confirmed to be caused by this PR.

apps/vscode/src/providers/format.ts
vezwork: I pulled down this PR and followed the reproduction steps in https://github.com/quarto-dev/quarto/issues/745 (which this PR should address). I ran into a new error that prevented any formatting from occuring:

Screenshot 2025-09-04 at 12 12 30 PM

I reverted this code change locally (bailing if edits are out of range) and formatting works again and the problem described in #745 is addressed.

juliasilge:
Interesting, no, I don't know why. I just tried it myself and I do in fact also see no extensions installed in the test runner VS Code that gets downloaded when you start such tests:

Screenshot 2025-09-05 at 3 17 23 PM

I wonder if this literally means "debug" rather than run:

> When you debug an extension test in VS Code

Like when you do _Test: Debug All Tests_ rather than run from the CLI, for example.

You might try adding the air and/or ruff extensions as extension dependencies just to see if you can get it installed in the test runner that way:

```json
  "extensionDependencies": [
    "posit.air-vscode",
    "charliermarsh.ruff"
  ]
```

Not that we want to really depend on these, but to explore what our options are.

juliasilge:
Another thought I had is that we might want to mock out a formatter to use in the tests, or maybe two (one that adds a final newline and one that does not).

vezwork:
I made a copy of this PR: https://github.com/quarto-dev/quarto/pull/818, it is
- rebased on main,
- with filtering of partial edits reverted,
- with an extension test that registers a custom formatting provider

Excuse me for copying the branch&PR to a new branch&PR, but I didn't feel comfortable force pushing (because of the rebase on main) to this one.

I don't currently know how to interpret the test results I am getting in #818: When I ran the test with the main fix by @lionel- reverted (the change in `codeForExecutableLanguageBlock`), the formatter does not seem to run at all... I don't understand why.

lionel-:
> I think the reason edits may be OOR is that virtualDocForCode() appends a newline at the end. That's probably a good idea because most formatters will add an empty line at end of document if not there, which we don't want in our chunks. Since there is an additional line in the vdoc, it's possible for the formatter to extend an edit up to that line, which doesn't exist in the original chunk. (We should add a comment about this in the code.)

Maybe a better way to go about this is to:

- Take note of how many empty lines a code chunk ends with.
- Don't append a newline in `virtualDocForCode()`, so that we format exactly the code chunk text.
- Apply edits as is. Most formatters will add a trailing newline if it doesn't exist.
- After applying, trim any trailing newlines, and insert the number of newlines we observed in the first step.

This way we don't need to adjust out of range text edits. This approach feels cleaner to me.

lionel-:
@vezwork Sounds good!

vezwork:
rebased on main and force pushed. For some reason I'm a co-author on the commits now.

vezwork:
Added a test that currently fails on this branch because of "out of range edits" from the mock formatter. It will also fail if the `quarto.formatCell` does not deal with formatters that do or don't append a newline in the same way such that if they otherwise produce the same formatting, the result will be the same because trailing newlines should be normalized (I think, based on my understanding of the discussion).

vezwork:
rebased on `main`

PR #755, opened 2025-07-07
Exclude section headers from workspace symbols in R packages

lionel-:
(This reopens https://github.com/quarto-dev/quarto/pull/751 after a weird auto-merge, sorry for any confusion.)

Branched from #750.
Addresses https://github.com/posit-dev/positron/issues/8401.

Once activated, the Quarto extension currently scans all md files and exports section headers to workspace symbols. While great generally, this can get in the way of the dev workflow in R packages.

R packages may include a variety of md files:

- Snapshot files: https://github.com/tidyverse/ellmer/tree/main/tests/testthat/_snaps
- Documentation: https://github.com/r-lib/rlang/tree/main/man/rmd

This causes Quarto to flood workspace symbols with headers that are mostly irrelevant for package development:

https://github.com/user-attachments/assets/a0269da3-68b5-44b2-b10c-ccae9a9196b3

I think by default (without any user configuration), headers shouldn't be exported as workspace symbols in R packages. But we need to preserve the existing behaviour for other projects.

To fix this, this PR:

- Introduces a new setting `quarto.symbols.exportToWorkspace`, with values `"default"`, `"all"`, or `"none"`. We need an enum with a default case instead of a boolean to provide the default behaviour that depends on the project type. The other two values allow hardcoding the behaviour in a project.

- When set to `"default"`, the LSP detects if the project is an R package.

  While the LSP would ideally be agnostic regarding the project type, having a customisation mechanism for extensions to set the default behaviour for a project would be complicated and require conflict solving if two extensions try to handle the same project.

- We already have a tool to detect R packages in the extension module (thanks to @juliasilge to have let me know!). It's a bit complicated because we need to parse the project's DESCRIPTION file if present, so I think we need to use the same implementation in both the LSP and extension. I pulled it up one level in `apps/utils`. To keep things light I didn't make it a full node module.

- The tool in `apps/utils` take a folder path. Then two wrappers in `apps/lsp` and `apps/vscode` are in charge of unpacking the relevant workspace folder (via different APIs specific to the LSP and extension) and calling the util.

### QA Notes

Create an R package with markdown files and make sure the Quarto extension is activated (it activates when navigating to an R file). Look for workspace symbols (cmd T, or `#` prefix in command palette). The markdown headers shouldn't be included.

In a project that's not an R package, they should be included.

In the R package you can opt into including the headers via `quarto.symbols.exportToWorkspace`. In the regular project you can disable them. The setting should immediately apply (at the next request).

### DIFF
--- a/apps/vscode/.vscode-test.mjs
+++ b/apps/vscode/.vscode-test.mjs
@@ -8,4 +8,13 @@ export default defineConfig([
       timeout: 5000,
     },
   },
+  // R project workspace
+  {
+    label: 'r-project',
+    files: 'test-out/r-project.test.js',
+    workspaceFolder: 'src/test/examples/r-project',
+    mocha: {
+      timeout: 5000,
+    },
+  },
 ]);
--- a/apps/vscode/.vscode/launch.json
+++ b/apps/vscode/.vscode/launch.json
@@ -13,13 +13,18 @@
       "outFiles": ["${workspaceFolder}/out/**/*.js"],
       "preLaunchTask": "${defaultBuildTask}"
     },
-    {
-      "type": "node",
-      "request": "attach",
-      "name": "Attach to Server",
-      "port": 6009,
-      "restart": true,
-      "outFiles": ["${workspaceRoot}/out/**/*.js"]
-    }
+		{
+			"name": "Extension Tests",
+			"type": "extensionHost",
+			"request": "launch",
+			"runtimeExecutable": "${execPath}",
+			"args": [
+				"--extensionDevelopmentPath=${workspaceFolder}",
+				"--extensionTestsPath=${workspaceFolder}/test-out",
+				"${workspaceFolder}/src/test/examples"
+			],
+			"outFiles": ["${workspaceFolder}/test-out/**/*.js"],
+      "preLaunchTask": "yarn: build-test"
+		}
   ]
 }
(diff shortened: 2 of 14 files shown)


juliasilge (commented):
This is working well, both in an R package and not!

Screenshot 2025-07-11 at 9 14 32 PM

Just two things here before you merge:

- Can you add a test for this PR to confirm that we still provide section headers for a regular `.qmd` file not in an R package? I set up the infrastructure for compiling and running tests in #757.
- Can you update the changelog for the extension?

lionel-:
@juliasilge We could add some extension tests for this but I think the LSP would be the more appropriate context for comprehensive tests?

juliasilge:
Comprehensive tests would be great, of course, but given the current state of this monorepo, we want to start where we are and have at least some discipline around adding testing with PRs. I think two good options for this PR would be either:

- add infrastructure for tests to `quarto-lsp`, including a GH action that runs them
- add a test to the VS Code extension confirming section headers, using the infrastructure set up in #757 

I leave it to your preference! What we want to avoid moving forward is to merge PRs here in this monorepo without any testing at all.

juliasilge (approved):
Thank you so much!

- This works great in R packages
- This works great in workspaces that are _not_ R packages
- I could run all the tests (new and old) via `yarn test-vscode`

juliasilge:
@vezwork Just a heads up that there are some additions to the testing infrastructure here, just for visibility for you.

PR #757, opened 2025-07-10
Add infrastructure for VS Code extension tests

juliasilge:
We want to get some first infrastructure set up here, so we can start writing extension level tests: https://code.visualstudio.com/api/working-with-extensions/testing-extension


juliasilge:
Now it's just `yarn test` from the working directory of the VS Code extension.

juliasilge:
I did it! 🎉 I have one very simple test, and tools to run that both locally and in CI.

cscheid (approved):
Very excited about this PR!

Everything looks good from what I can tell. It also seems very low risk to just merge this, given it doesn't really interact with the other parts of the code base. If the CI action acts up, we can just disable the workflow.

PR #770, opened 2025-07-22
Add tests and test support commands for visual editor

vezwork:
image

---

This PR builds on-top of https://github.com/quarto-dev/quarto/pull/757 to add tests, supporting functionality in test-utils, and registers visual editor commands specifically for tests. Two tests were added, they test: 
1. opening the visual editor, and 
2. roundtripping a qmd file through the visual editor.

Unfortunately these tests rely on `wait`ing an approximate amount of time for the visual editor to open and for other related commands to complete. They are slow and potentially flaky. I think this could be improved, but I'm not yet sure how. We may be able to return things from commands that help us understand if the visual editor has finished opening (maybe we will have to do some polling?). I wonder how other teams do this...

The tests pass on my machine and its kind of fun to watch :)

edit: I just noticed [the tests fail](https://github.com/quarto-dev/quarto/actions/runs/16452713465/job/46502077104?pr=770) on this PR. It seems to be because the extension is not built before the tests are ran, resulting in the commands that were added in this PR not being present at the time of running tests. I attempted a fix (in now-reverted commits) that:
- ran the `test-extension` job after the `package-extension` job but that did not help;
- I also tried putting the test workflow steps at the end of the `package-extension` job, but that did not help;
- I also tried registering the added testing-specific commands into the extension's package.json, but that did not help.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,6 +25,12 @@ jobs:
           sudo apt-get -y update
           sudo apt-get -y install --fix-missing xvfb
 
+      - name: Build vscode extension
+        run: |
+          cd apps/vscode
+          yarn install
+          yarn run build
+
       - name: Compile and run tests
         run: |
           yarn install --immutable --immutable-cache --check-cache
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -25,6 +25,8 @@ jobs:
           sudo apt-get -y update
           sudo apt-get -y install --fix-missing xvfb
 
+      - uses: quarto-dev/quarto-actions/setup@v2
+
       - name: Build vscode extension
         run: |
           cd apps/vscode
--- a/apps/lsp/src/index.ts
+++ b/apps/lsp/src/index.ts
@@ -29,9 +29,9 @@ import {
   WorkspaceSymbol
 } from "vscode-languageserver";
 
-import { CompletionItem, Hover, Location } from "vscode-languageserver-types"
+import { CompletionItem, Hover, Location } from "vscode-languageserver-types";
 
-import { createConnection } from "vscode-languageserver/node"
+import { createConnection } from "vscode-languageserver/node";
 
 import { URI } from "vscode-uri";
 import { TextDocument } from "vscode-languageserver-textdocument";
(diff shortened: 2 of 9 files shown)


cscheid:
I'm pretty out of my depth here, but it looks like `activateEditor` (`quarto/apps/vscode/src/providers/editor/editor.ts`) isn't getting called, or those ids are not being set as one would expect. Is there a way to test this hypothesis?

vezwork:
I was able to verify that the extension's commands are not present at all. I removed the `runCommand`s for the commands I added in this PR and only left in `runCommand`s for existing Quarto commands. Here's what I got:

```
 Error: command 'quarto.editInVisualMode' not found
```

I'm gonna check out what @juliasilge sent! That second link looks really promising for ensuring the extension is activated.

vezwork:
Some progress. I tried to explicitly activate the extension in the tests and got this error in the logs:
```
 Activating extension 'quarto.quarto' failed: Cannot find module '/home/runner/work/quarto/quarto/apps/vscode/out/main.js'
 ```

**Comment by vezwork** [comment]
Ok dang... I added a step to `cd app/vscode`, `yarn install` and `yarn run build` to the test workflow. Activating the extension no longer fails, but we're back to this error:
```
Error: command 'quarto.test_setkVisualModeConfirmedTrue' not found
```

juliasilge:
Ah, of course! We should definitely install Quarto in this workflow! I definitely should have done that already 🙈 

I think a good place to start is just the regular install that you can do via:
https://github.com/quarto-dev/quarto-actions

vezwork:
@lionel- I am getting a bit stuck on debugging the quarto extension test github workflow. Do you have any advice or guidance?

vezwork:
Skipping the added in tests in the workflow/CI for now. This PR is ready for review and merge. We can come back to getting the extension, more specifically the LSP, working in CI later.

lionel-:
@juliasilge 

> I notice that in air, you do this to wait for the LSP to be fully started up in tests:
>
> https://github.com/posit-dev/air/blob/7125f07fb88ea9b79387b04d0a8fa131b3cc5044/editors/code/src/test/extension.ts#L46
> 
> Could you say more about that in a testing context? And how we might apply that to the Quarto extension?

This starts the LSP and only resolves when fully started (the heavy lifting is done by vscode-languageclient's `start()` method). It also ensures the quarto extension has been activated, as relying on activation events can be flaky. Tests that rely on LSP features should generally call something like this on init to ensure proper synchronisation.

However from the error messages above it looks like the LSP is getting started but fails to do so.

PR #777, opened 2025-07-30
Debug: log to show difference in `vscode.executeCompletionItemProvider` in vscode vs Positron

vezwork:
Please see the comments on the modified code in this PR.

### DIFF
--- a/apps/vscode/src/providers/editor/codeview.ts
+++ b/apps/vscode/src/providers/editor/codeview.ts
@@ -105,12 +105,12 @@ export function vscodeCodeViewServer(_engine: MarkdownEngine, document: TextDocu
       // if this is Positron, no visual editor completions
       // TODO: fix LSP issues for visual editor in Positron:
       // https://github.com/posit-dev/positron/issues/1805
-      if (hasHooks()) {
-        return {
-          items: [],
-          isIncomplete: false
-        };
-      }
+      // if (hasHooks()) {
+      //   return {
+      //     items: [],
+      //     isIncomplete: false
+      //   };
+      // }
 
       // otherwise delegate to vscode completion system
       const vdoc = virtualDocForCode(context.code, language);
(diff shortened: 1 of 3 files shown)


apps/vscode/src/providers/editor/codeview.ts
```diff
@@ -105,12 +105,12 @@ export function vscodeCodeViewServer(_engine: MarkdownEngine, document: TextDocu
...
-          items: [],
-          isIncomplete: false
-        };
-      }
+      // if (hasHooks()) {
+      //   return {
+      //     items: [],
+      //     isIncomplete: false
+      //   };
+      // }
```
vezwork, line 113: For debugging purposes: re-enable completions in the visual editor in Positron

packages/editor-codemirror/src/behaviors/completion.ts
```diff
@@ -114,6 +114,8 @@ async function getCompletions(
     return null;
   }
 
+  console.log('getCompletions', completions, [...completions.items])
+
```
vezwork, line 118: In vscode, in the extension host, in the developer console: Image

In Positron, in the extension host, in the developer console: 
Image

juliasilge:
You might want to change your VS Code setup to use Jedi instead of Pylance for Python, for a better comparison.

vezwork:
After going over all this information with @juliasilge (also see https://github.com/posit-dev/ark/issues/900), we have arrived at a plan for a solution to the problem of completions not being properly sorted and filtered in the VE in Positron (we now know it affects Jedi users in VSCode as well).

I will try copying&pasting from what seems to be [the code where the source editor sorts completions](https://github.com/posit-dev/positron/blob/a6c2109d7d5770e898d1ac147085843841bcd7d9/src/vs/editor/contrib/suggest/browser/completionModel.ts#L140). This will give consistency between the source editor and the visual editor; it will respect the LSP spec which says completion sorting and filtering should be handled by the client; it adds a maintenance burden (if the source editor sorting/filtering does change and we the visual editor to remain consistent with it).

PR #778, opened 2025-08-04
don't prepend delimiter to path

cscheid:
(thanks @jforest, who spotted this in the wild) I think we shouldn't be prepending `path.delimiter + ...` to the path. We end up with a path string starting with `path.delimiter`, which puts an empty string on the path.


juliasilge (commented):
Great catch, @jforest!

Things are working well for me with this change. Can you add an entry to the `CHANGELOG`? 

Once this is merged in, I can do a release of the extension if we want this fix out ASAP.

cscheid:
(oops, we both tried adding to changelog. Sorry! Though I'll keep mine because I think it's more descriptive.)

PR #781, opened 2025-08-07
add initial snapshot tests

vezwork:
## Description

- Adds a `readOrCreateSnapshot` helper function that makes snapshot tests convenient to create:
  - If you don't already have a snapshot, it creates one with the passed contents and returns the contents
  - If you do already have a snapshot, it reads it and returns the contents
- Adds two roundtripping snapshot tests based on Quarto documents from https://github.com/quarto-dev/quarto-markdown-test-corpus
  - `valid-basics.qmd`
  - `invalid.qmd`

### For fun, here's a screenshot of the diff between `valid-basics.qmd` before and after roundtrip
I got this in VSCode by right clicking on `valid-basics.qmd`, then clicking "select for compare", then right clicking `generated_snapshots/roundtripped-valid-basics.qmd`, then clicking "compare with selected".
image


juliasilge (approved):
This works great for me locally as well!

PR #798, opened 2025-08-22
improve completions sorting

vezwork:
I was originally planning to copy completions sorting code from VSCode, but it is incredibly complex and cannot be directly copied&pasted (because we probably don't want to rewrite CodeMirror's Autocompletion from scratch). If I was going to copy VSCode's completion code, at the very least I would first have to understand how CodeMirror Autocompletion API works... 

So, in this PR I have 
- refactored the existing VE's completion sorting and filtering code for readability, 
- manually improved the VE's completions sorting in Positron and VSCode with the Python Jedi language server, and
- accept completion with <kbd>Tab</kbd>.

Upon initial testing, the completions are good in both Python and R in Positron and VSCode with the Python Jedi language server. It may be best to go with this simple fix for now. In the future it would be good if we can understand more what is happening in VSCode completions, but I have realized that could be quite a bit of work outside of my domain.

### DIFF
--- a/apps/vscode/src/providers/editor/codeview.ts
+++ b/apps/vscode/src/providers/editor/codeview.ts
@@ -102,16 +102,6 @@ export function vscodeCodeViewServer(_engine: MarkdownEngine, document: TextDocu
         return lspCellYamlOptionsCompletions(context, lspRequest);
       }
 
-      // if this is Positron, no visual editor completions
-      // TODO: fix LSP issues for visual editor in Positron:
-      // https://github.com/posit-dev/positron/issues/1805
-      // if (hasHooks()) {
-      //   return {
-      //     items: [],
-      //     isIncomplete: false
-      //   };
-      // }
-
       // otherwise delegate to vscode completion system
       const vdoc = virtualDocForCode(context.code, language);
       const completions = await vdocCompletions(


vezwork:
> I do notice a few difference in R completions, but I am not sure with my current level of review whether we should change what happens in ark or here in this logic.

I am not sure either. If we can identify issues with R completions, I would now say it shouldn't be difficult to address them here. 

Tangential discussion, but: note that since ark always gives like 2500+ completions, and since sorting and filtering is done in the UI process, processing completions is consistently slightly slow in R code and makes the UI feel slightly less responsive. I might try moving a bunch of the filtering to the LSP so its not in the UI process. I feel like ideally ark would do some basic filtering on their side(such as inclusion of what the user is typing in the proposed completions) for performance.

> Just in case you might already know, an immediate noticeable difference I am seeing is that R packages show up in the visual editor but not the source editor:
> [..]
> It's the `bslib` and `jquerylib` entries there.

I noticed these! I didn't know what they were but I played around with them and they seem useful?

vezwork:
I wonder if I should just do another quick pass over this then go to merge it. I can't think of another step forward. Maybe I should get someone who uses R in Positron to try it out.

vezwork:
Thank you @juliasilge, totally clarifying.

vezwork:
I have uncommitted code that fixes position miscalculations that result in snippets often being filtered out of completions. After I fixed that, I found out that snippets with nested fields are somewhat broken in codemirror's autocomplete extension. I raised an issue in their repo here: https://github.com/codemirror/dev/issues/1607

vezwork:
@juliasilge I have added 

- filtering for namespaces in completions (`bslib` and the like no longer show up), and
- Tab-to-accept-completion support.

The PR is ready!

apps/vscode/src/providers/editor/codeview.ts
```diff
@@ -105,12 +105,12 @@ export function vscodeCodeViewServer(_engine: MarkdownEngine, document: TextDocu
...
-          items: [],
-          isIncomplete: false
-        };
-      }
+      // if (hasHooks()) {
+      //   return {
+      //     items: [],
+      //     isIncomplete: false
+      //   };
+      // }
```
juliasilge, line 113: Let's remove this code before we merge!

juliasilge (approved):
This is such an incredible improvement! Having <kbd>Tab</kbd> set up to accept completions is such a relief. 😅  I think we should merge and move forward! 

I do think we're going to get reports of differences in the completions in visual vs. editor mode, but we can log those and address them moving forward. Maybe we can even set up some testing infrastructure so we can test completions for the visual editor, to specify the expected behavior via tests? Like unit tests for the `editor-codemirror` package here?

@DavisVaughan Thank you for outlining those thoughts. Should we log any of that additional detail at https://github.com/posit-dev/ark/issues/900?

packages/editor-codemirror/src/behaviors/completion.ts
```diff
@@ -102,135 +102,163 @@ export function completionBehavior(behaviorContext: BehaviorContext): Behavior {
...
+  // Note: in contrast, when you type `os.a` replaceText will give `a` for a completion like `abc`.
+  const token = context.matchBefore(/\S+/)?.text;
 
-        // only allow non-text edits if we have no token
-        if (!item.textEdit && token) {
-          return false;
-        }
+  const filteredItems = items.filter(item => {
+    // no namespaces
+    if (vsKindToType(item.kind) === 'namespace') return false;
```
DavisVaughan, line 229: I would strongly recommend not wholesale removing namespaces

Namespaces are included in both RStudio and Positron's console and it is very important to be able to complete these

Screenshot 2025-09-03 at 10 28 33 AM
Screenshot 2025-08-20 at 8 58 26 AM

@juliasilge had made a comment about bslib not showing up when she typed `lib` in source mode, but I think that just has to do with how VS Code itself filters based on the _text_ and doesn't have anything to do with the _type_ of completion in question.

For example, if I do the same thing i see lightgbm pop up in the positron console, which suggests that namespace completions are definitely still there

Screenshot 2025-09-03 at 10 32 19 AM


So I vote we just remove this because keeping it would make for a worse experience overall

juliasilge: I guess I am OK with that, and we can log a new issue to address the specific filtering that happens in this type of case. I do think we need to get some unit tests set up for completions so we can specify the behavior we expect in a fairly low-burden way, and then can iterate moving forward.

DavisVaughan: @juliasilge FWIW if we do https://github.com/quarto-dev/quarto/pull/798#issuecomment-3249168617 then your experience here would be massively improved because only things starting with `dpl` would be provided by ark in the initial set of completions

vezwork: I'll remove the filtering of namespaces for now!

vezwork: @juliasilge I made an issue to capture the differences in filtering in the case of namespaces https://github.com/quarto-dev/quarto/issues/808

DavisVaughan: thanks @vezwork!

PR #801, opened 2025-08-29
fix snippet completion incorrect filtering

vezwork:
## Problem

Snippets were being filtered from completions because of a position miscalculation. 

The `itemFrom` function was not correct, it was doing something clever that it is not quite right.

`itemFrom` should calculate how many characters into the document the start of a completion is. It did this by calculating the length of the completion and subtracting that from the text cursor's position. The way it was calculating the length of the completion was not correct. The completion gives its positions in the context of the `cvContext.code`, which is different from the positions in the actual document; For one thing, the line containing the language string (like `{r}`) is removed from `cvContext.code`. So, we can't use the positions the completion gives us directly. That is why `itemFrom` was cleverly not using the positions, but instead using the positions to calculate the length of the completion. 

But, for example, if you are typing `libr`, a completion for the `library` snippet will say that its insertion character position is from 0 to 7. What's up with that? For whatever reason, since your cursor is at the end of a line with 4 characters, the completion pretends that it can replace 3 more non-existent characters in the rest of the line. If there were any characters in the line after `libr`, even just whitespace like `libr `, then the completion would instead say that its insertion character position is from 0 to 4.

This PR fixes the issue by snapping the ending position of a completion to the end of its line in `cvContext.code`.

After I fixed this issue, I was able to easily insert the `library` snippet in R code, and this uncovered a bug in snippets in codemirror autocomplete that I reported here https://github.com/codemirror/dev/issues/1607.


vezwork:
@marijnh fixed the codemirror/autocomplete [issue](https://github.com/codemirror/dev/issues/1607) and made a [release of it](https://github.com/codemirror/autocomplete/releases/tag/6.18.7). I spent a couple hours this morning attempting to upgrade our codemirror/autocomplete version. I was not successful (facing Quarto monorepo problems that I don't understand).

I feel that this PR is blocked on getting this autocomplete fix in. If we are going to fix it so that snippet completions show up properly, snippet insertion should work.

I will have to get help figuring out how to upgrade the package.

cscheid:
I imagine part of what's happening here is that this is a https://turborepo.com/ monorepo, and yarn might be getting extra confused in the process?

(I would happily switch us all away to better tooling if the benefits are there, but I would really like to not to have to switch again in a year. So I think adding some friction here and thinking half a dozen times before pulling the trigger is worthwhile.)

vezwork:
@cscheid  It seems that we would have to embark on a proper migration of our monorepo to a new version of yarn to deal with these issues. It scares me though. I have no sense for how much work that would be.

vezwork:
Also, the yarn upgrade causes the number of changes in this PR to be wild. It should be really be in its own PR. I'm hesitant to start such a PR until we decide we actually want to invest into a proper migration.

vezwork:
I checked out the Positron repo to see how they are managing their dependencies and they are also using an old version of yarn...

An alternative to thing to try rather than upgrading yarn: I heard there is something we can maybe do with the old version of yarn to do with "no hoisting" that could possibly help.

PR #825, opened 2025-09-24
use `statementRangeProvider` in Positron to execute statements at cursor in VE

vezwork:
New in Positron: pressing ctrl+enter with your cursor in a multiline R statement executes the whole statement.

Screenshot 2025-09-24 at 3 55 59 PM

Before it would only execute the current line.

Note: this fix relies on Positron's `vscode.executeStatementRangeProvider` command and does not work in VSCode. I was attempting to get this working in VSCode by creating a virtual doc for the code block, setting your cursor in the virtual doc, and running a cell executor on the virtual doc. That did not work out because I don't know how to set your cursor's position in a virtual doc and cell execution commands don't take positions as arguments.

### DIFF
--- a/apps/vscode/src/providers/cell/commands.ts
+++ b/apps/vscode/src/providers/cell/commands.ts
@@ -49,6 +49,10 @@ import { ExtensionHost } from "../../host";
 import { hasHooks } from "../../host/hooks";
 import { isKnitrDocument } from "../../host/executors";
 import { commands } from "vscode";
+import { virtualDocForCode, withVirtualDocUri } from "../../vdoc/vdoc";
+import { embeddedLanguage } from "../../vdoc/languages";
+import { Uri } from "vscode";
+import { StatementRange } from "positron";
 
 export function cellCommands(host: ExtensionHost, engine: MarkdownEngine): Command[] {
   return [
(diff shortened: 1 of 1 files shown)


cscheid:
I think it's ok to prioritize Positron fixes and leave the VSC fix open for a community PR if there's an approach there that would work.

juliasilge:
This is really exciting to see! 🎉 

I think there is a bit of improvement to do in terms of where the cursor goes in the visual editor after using these. I see that you said there is some difficulty about setting the cursor position in the visual editor 😬 so it's possible this is a big challenge. Take a look at what I mean:

````
---
title: "Diamond sizes"
format: html
---

```{r}
library(tidyverse)

smaller <- diamonds |>
  filter(carat <= 2.5)
```

````

- In the source editor, put your cursor at `library`
- Use <kbd>Cmd+Enter</kbd> to step through the lines, including the multiline R statement (notice where your cursor goes _after_ each statement execution
- Switch to the visual editor and put your cursor back at `library`
- Use <kbd>Cmd+Enter</kbd> to try to step through the lines (notice that the cursor doesn't advance enough, which feel especially weird on the empty line)

I also wanted to ask about Python, because I don't see the Python statement range provider being applied, I think? Here is an example to look at:

````
---
title: "Quarto Basics"
format: html
---

## Matplotlib

For a demonstration of a line plot on a polar axis, see @fig-polar.

```{python}
#| label: fig-polar
#| fig-cap: "A line plot on a polar axis"

import numpy as np
import matplotlib.pyplot as plt

r = np.arange(0, 2, 0.01)
theta = 2 * np.pi * r
fig, ax = plt.subplots(
    subplot_kw = {'projection': 'polar'}
    )
ax.plot(theta, r)
ax.set_rticks([0.5, 1, 1.5, 2])
ax.grid(True)
plt.show()
```

````

- In the source editor, put your cursor at `import numpy`
- Use <kbd>Cmd+Enter</kbd> to step through the lines, including the multiline Python statement (notice where your cursor goes _after_ each statement execution
- Switch to the visual editor and pur your cursor back at `import numpy`
- Use <kbd>Cmd+Enter</kbd> to try to step through the lines (notice it executes the whole cell, not line by line)

juliasilge:
I think this was only applying to R documents because of this here:

https://github.com/quarto-dev/quarto/blob/2dede3f9306d8a91913852d35bcfba170d270b80/apps/vscode/src/providers/cell/commands.ts#L344

--------

What happens in the source editor is something like this:

- Positron says, "I have a `quarto` document and I need a statement range provider for it"
- Quarto extension says, "I have one of those":

https://github.com/quarto-dev/quarto/blob/012d457eeda5a00d8a9d97b2af0c692cefdbd02c/apps/vscode/src/host/hooks.ts#L97

- Positron says, "OK, I want you to give me the statement range at position X"
- Quarto extension say, "I will do that":

https://github.com/quarto-dev/quarto/blob/012d457eeda5a00d8a9d97b2af0c692cefdbd02c/apps/vscode/src/host/hooks.ts#L161

- Quarto extension fulfills `getStatementRange()` by executing `vscode.executeStatementRangeProvider` for the virtual doc (i.e. it _asks Positron_ for the statement range for the underlying virtual doc, since Positron has statement range providers for Python and R):

https://github.com/quarto-dev/quarto/blob/012d457eeda5a00d8a9d97b2af0c692cefdbd02c/apps/vscode/src/host/hooks.ts#L186

So it goes back and forth a fair amount!

I _think_ you will want to set up request forwarding in the `visualEditorServer` function in `connection.ts`, in `codeViewServerMethods`; I think what needs to happen for a bunch of the LSP methods (for the statement range provider, <kbd>F1</kbd> help, signature help, etc) is request forwarding. My understanding is that piping through this LSP functionality new to the visual editor will be similar to how completions are piped through.

vezwork:
Statement execution (with cursor advancement to the next statement) should now be working for both R and Python cells in the Visual Editor in Positron.

A caveat:
- there is a hard-coded assumption in index calculation code about what the first line of a code cell looks like, its assumed it looks like `{r}` or `{python}` depending on the language.
  - right now the first line is not passed around with the code block data, which is why I didn't use the actual contents
  - would the first line ever not look like this?

cscheid:
> A caveat:
> 
>     * there is a hard-coded assumption in index calculation code about what the first line of a code cell looks like, its assumed it looks like `{r}` or `{python}` depending on the language.
>       
>       * right now the first line is not passed around with the code block data, which is why I didn't use the actual contents
>       * would the first line ever not look like this?

I think this is a safe assumption. Someone might one day want to run code cells with extra braces, like `{{r}}`. But those are not actually executed by Quarto, so we're ok.

cscheid (approved):
One minor comment on type design, but otherwise LGTM.

Issue #687, opened 2025-04-01
Ctrl+Enter fails to send code inside `sh`/`bash` chunks to terminal


sharon-wang:
### Bug description

<kbd>Ctrl+Enter</kbd> and <kbd>Run Cell</kbd> fail to send code inside `sh` chunks to terminal in both VS Code and Positron.

Copied over from @Robinlovelace's issue in Positron: https://github.com/posit-dev/positron/issues/7047.

Appears to be a regression of: https://github.com/quarto-dev/quarto/issues/4.

### Steps to reproduce

Create a code chunk like this:

````
```{sh}
ls
```
````

Place your cursor on the `ls` bit.

Keyboard input <kbd>Ctrl+Enter</kbd> or click <kbd>Run Cell</kbd> in the IDE

### Actual behavior

Nothing appears to happen. No code is sent to the Terminal.

### Expected behavior

The code should be sent to the terminal and executed as shown below (entered manually):

![Image](https://github.com/user-attachments/assets/2af728e7-6e61-4eaa-a806-d12cecb6d8d6)

### Your environment

Ubuntu + Positron

```
Positron Version: 2025.03.0 build 116
Code - OSS Version: 1.96.0
Commit: 7f0f93873e8b24c9f393bb3586724f4783c5e72c
Date: 2025-03-01T18:48:08.166Z
Electron: 32.2.6
Chromium: 128.0.6613.186
Node.js: 20.18.1
V8: 12.8.374.38-electron.0
OS: Linux x64 6.8.0-52-generic
```

Mac 15.3.2 + Positron

```
Positron Version: 2025.04.0 (Universal) build 204
Code - OSS Version: 1.98.0
Commit: cfac68acb45f7646202711f64f46643f51c0ba2f
Date: 2025-03-27T03:36:09.040Z
Electron: 34.2.0
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

Quarto Version: 1.119.0 (2025-03-27)
```

Mac 15.3.2 + VS Code

```
Version: 1.98.2
Commit: ddc367ed5c8936efe395cffeec279b04ffd7db78
Date: 2025-03-12T13:32:45.399Z
Electron: 34.2.0
ElectronBuildId: 11161602
Chromium: 132.0.6834.196
Node.js: 20.18.2
V8: 13.2.152.36-electron.0
OS: Darwin arm64 24.3.0

[...]

Issue #4, opened 2022-11-30
Ctrl+Enter executes R and Python code but not bash


Robinlovelace:
As shown below: 

![image](https://user-images.githubusercontent.com/1825120/204927556-1e1a79a8-57ff-489b-82de-ad4cade6a34d.png)

Any way to get it to execute bash commands?

jjallaire:
We've just added support for "Run Cell" for bash: https://github.com/quarto-dev/quarto/commit/250a98441c403c25630f61a1234290273ff2877d

Robinlovelace:
Awesome. Thanks! Shall I give it a test run?

jjallaire:
Yes, we just pushed an update to the marketplace w/ this (v1.63.0)

Robinlovelace:
![Peek 2023-02-10 00-05](https://user-images.githubusercontent.com/1825120/217967188-6fe84d00-d402-4c50-88bf-76bccfec6e98.gif)

Confirmed. Demo above. Many thanks!

gortegasolis:
Hi, 
Bash is not working for me. I tried adding "engine: knitr" and adding first an R chunk but nothing works.

This is my system:

Version: 1.89.1
Commit: dc96b837cf6bb4af9cd736aa3af08cf8279f7685
Date: 2024-05-07T05:16:23.416Z
Electron: 28.2.8
ElectronBuildId: 27744544
Chromium: 120.0.6099.291
Node.js: 18.18.2
V8: 12.0.267.19-electron.0
OS: Linux x64 6.5.0-35-generic

mdehollander:
Same here, doesn't seem to work (anymore)

Robinlovelace:
Worth testing and reopenning @jjallaire and team?

sharon-wang:
For those following along, new issue opened here: https://github.com/quarto-dev/quarto/issues/687

PR #826, opened 2025-09-24
Fix cell executor selection logic so bash cell execution works

vezwork:
Fixes https://github.com/quarto-dev/quarto/issues/687

There was a bug in the logic that checked if a cell executor required an extension to run. As a result, bash cells could not be run. That is fixed, and you can send bash/sh cells to the console now. Also added some logic to avoid sending bash code cells to terminals running R language interactive sessions.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed Copilot completions in `.qmd` documents (<https://github.com/quarto-dev/quarto/pull/887>).
 - Fixed a bug where the `autoDetectColorScheme` setting could cause equation previews to have a dark text on dark background and vice versa (<https://github.com/quarto-dev/quarto/pull/864>).
+- Fix a regression where bash cell execution does not work (<https://github.com/quarto-dev/quarto/pull/826>).
 
 ## 1.128.0 (Release on 2026-01-08)
 


cscheid:
Am I missing something or should this ensure that the terminal is running bash? There's probably at least one sicko out there running powershell on macos as their main shell. (And will this work on windows where... isn't powershell is the default?)

cderv:
As this is sending to terminal, it won't matter which is the terminal. So 
```{bash}
ls
```

will send the command to the active terminal. In my case this is powershell. `ls` is a known alias for powershell equivalent, and so it works. 

But if it were bash only command, then this would not. 

To clarify, this implementation is different than running the cell through quarto, with engine **knitr** for example. 

A bash chunk in a .qmd / .Rmd documents, requires on windows to have bash emulator (like Git bash) and configure **knitr** to use it. So usually a Windows user would not have bash chunk. 

For Jupyter engine, I guess this would be the same. 

Here is the working part when running cell, and error we see on Windows when rendering with Quarto
Positron_AdFikrlHXp


So this PR fixes the issue, of the initial feature that was introduced in 
- https://github.com/quarto-dev/quarto/issues/4

However, ultimately, this type of feature raises the question of the expectation for user between running interactively and rendering. 

But you know the discussion... 

Sorry I missed that PR - happy to help on windows stuff always!

cscheid:
> But if it were bash only command, then this would not.

I think this is what worries me. It's a "bash" cell. It should work with bash-only commands.

cderv:
Currently, it is a convenient feature that assumes users running .qmd cell interactively have the same terminal as the _terminal_ cell they are using... 😅 

Not sure if there is a way to check the terminal that is open, and compare to the name of the cell language.

Issue #829, opened 2025-09-26
Outline includes headers in divs


vezwork:
~A regression of https://github.com/quarto-dev/quarto-cli/issues/716.~ @kmasiello mentioned that this had regressed in the Quarto slack ([Posit internal slack link to discussion](https://positpbc.slack.com/archives/C01LQETLA7L/p1758893689539949)).

> [!NOTE]  
> Edit: ~I've realized that technically this is not a regression~. In [JJ's original implementation](https://github.com/quarto-dev/quarto-vscode/commit/5e9e20f67a2a4da07ee9a391c4d25ac2a486e2a6), only those headers which are inside callouts (divs with classname starting with callout-) were excluded. ~That is currently working.~ edit 2: that is actually not totally working. See subtle repro example at bottom of this issue.
>
> However, it does seem that the outline should match the TOC that ends up in the rendered document. I believe that the TOC does not include headers inside ANY div. This issue describes that.

The outline should not include headers inside of divs / fences. This matches the TOC behaviour when rendering documents.

### Incorrect (what is currently happening)
<img width="622" height="179" alt="Image" src="https://github.com/user-attachments/assets/2e9f48d6-4400-4460-9e82-f34a6ac47b08" />

note: I realized after taking this screenshot and reading the relevant code that if the classname on the callout started with `callout-` (note the trailing dash!) and not just `callout`, it would've been excluded from the outline.

### Correct (what should happen)
<img width="619" height="180" alt="Image" src="https://github.com/user-attachments/assets/f14bb3e7-b298-4362-a7c7-073157106c53" />

## Repro example qmd:
```
---
title: "My document"
format: "html"
---
## Level 2 Title 1

### Level 3 Title 1

::: {.example-non-callout-div}
## Level 2 Title 1 - inside div
:::

### Level 3 Title 2
```

## More subtle repro example edited down from what @kmasiello gave me:
```
::: {.callout-note}

## I should not be in the Outline!

when this text is here (on the line before the triple colon to end the div), this shows up in the outline and there is no div in the list of document tokens!?
:::
```

vezwork:
@cscheid yes please

PR #830, opened 2025-09-26
Exclude headers in divs from outline

vezwork:
Fixes https://github.com/quarto-dev/quarto/issues/829

Some headers inside divs were already being excluded, but the exclusion rules were too conservative. The exclusions checked if the header was inside a callout (a div with a class that starts with `callout-`), a Theorem (a div with a theorem-specific id), or a Proof (a div with a proof-specific id). Now it excludes all headers inside any div.

### DIFF
--- a/apps/lsp/src/service/toc.ts
+++ b/apps/lsp/src/service/toc.ts
@@ -36,6 +36,7 @@ import {
   getLine,
   Document,
   Parser,
+  isDiv,
 
 } from 'quarto-core';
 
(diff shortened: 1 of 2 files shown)


apps/lsp/src/service/toc.ts
```diff
@@ -140,7 +141,7 @@ export class TableOfContents {
     }
 
     // compute restricted ranges (ignore headings in these ranges)
-    const isWithinIgnoredRange = isWithinRange(tokens, token => isCallout(token) || isTheorem(token) || isProof(token));
+    const isWithinIgnoredRange = isWithinRange(tokens, isDiv);
```
cscheid, line 144: Let me make things fun (of the type 3 variety).

What should we do about conditional content divs? (to say nothing of arbitrary Lua filters...)

cscheid: The behavior of conditional content divs is that the div itself will never actually appear in the resulting document. But if the conditions are met, then the contents will be included in the document, and headers inside will "move one level up".

cscheid: @vezwork No, I don't think that's the best move. Let's merge this as is and talk about a redesign for later.

vezwork:
Closing this PR, it is not a proper fix of the issue.

Issue #819, opened 2025-09-11
Cleanup issues with old .vdocs


lgaborini:
### Bug description

I'm using Positron with Quarto extension 1.124.0, this issue might be related to #708 (confirmed fixed).

The former syntax errors keep piling up under the Problems panel, even if the files in the Temp directory (or the directory itself) are no longer there:

<img width="1337" height="770" alt="Image" src="https://github.com/user-attachments/assets/4316af64-5144-46a0-bbd1-76acffe46139" />

(I'm editing a fairly large Quarto project with many files and Positron is getting sluggish. Not sure if this is the culprit!)

### Expected behavior

Empty Problems window if the project contains no syntax error

### Your environment

Positron Version: 2025.09.0 (user setup) build 139
Code - OSS Version: 1.103.0
Commit: 30e3d0a74e09a0f6640929822cb57a326671360e
Date: 2025-08-29T18:50:28.998Z
Electron: 37.2.3
Chromium: 138.0.7204.100
Node.js: 22.17.0
V8: 13.8.500258-electron.0
OS: Windows_NT x64 10.0.19045

Quarto extension: 1.124.0

lgaborini:
I agree! Especially given that these files no longer exist in the filesystem, but one can still open them.
It's already a miracle that I don't get more errors such as locked files due to Onedrive/Dropbox or shared drives...

I just hope you can find a way to make diagnostics work otherwise: maybe use one file for everything, just to fill up the panel?

PR #832, opened 2025-09-30
Hide diagnostics from virtual documents

juliasilge:
Addresses https://github.com/quarto-dev/quarto/issues/819, which has been quite a bear in real work! Many, many diagnostics from our `vdoc` files, every time you save a file with a syntax error in it. 😱 

It turns out the `Middleware` from "vscode-languageclient" already has a way to do `HandleDiagnosticsSignature`, so this was fairly straightforward.

I will get started on an extension test for this.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -1045,6 +1045,13 @@
           "default": true,
           "markdownDescription": "Use reticulate to execute Python cells within Knitr engine documents."
         },
+        "quarto.vdoc.hideDiagnostics": {
+          "order": 26,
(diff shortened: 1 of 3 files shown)


juliasilge:
@apsteinmetz I believe you are seeing https://github.com/quarto-dev/quarto/issues/855 which we addressed in https://github.com/posit-dev/positron/pull/11225. That will get into a regular Positron release at the beginning of February, or you can get the fix now if you switch to the [dailies](https://positron.posit.co/updating.html#daily-builds).

PR #840, opened 2025-10-05
Add `claude.md` with info on the repo structure, to help Claude know where things are

juliasilge:
If I ever use Claude Code in this repo, I have to spend time telling Claude where things are or have it spend tokens finding out itself. Maybe we can try this out? On some initial exploration it seems pretty helpful.

### DIFF
--- a/claude.md
+++ b/claude.md
@@ -1,0 +1,35 @@
+# Quarto Development Guide
+
+## Project Overview
+
+Quarto is an open-source scientific and technical publishing system built on [Pandoc](https://pandoc.org). This repository contains the source code for various parts of the Quarto ecosystem, with the main CLI implementation housed in a separate repository ([quarto-cli](https://github.com/quarto-dev/quarto-cli)).
+
+### Main Components
+
+- **VS Code Extension**: The primary VS Code extension for working with Quarto documents
+- **Writer**: A web-based editor for Quarto documents
+- **LSP**: Language server for Quarto documents
+- **Core Packages**: Shared libraries used across multiple components
+
+## Repository Structure
+
+The repository is organized as a monorepo using Yarn workspaces:
+
+- `apps/`: Contains standalone applications
+  - `vscode/`: VS Code extension for Quarto
+  - `writer/`: Web-based Quarto editor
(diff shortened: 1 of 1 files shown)


claude.md
```diff
@@ -0,0 +1,35 @@
...
+### Main Components
+
+- **VS Code Extension**: The primary VS Code extension for working with Quarto documents
+- **Writer**: A web-based editor for Quarto documents
+- **LSP**: Language server for Quarto documents
+- **Core Packages**: Shared libraries used across multiple components
+
+## Repository Structure
+
+The repository is organized as a monorepo using Yarn workspaces:
```
cscheid, line 16: I would mention [turbo](https://turborepo.com/) because, AIUI, it changes where node_modules go and how monorepo dependencies are declared in package.json files.

claude.md
```diff
@@ -0,0 +1,35 @@
...
+
+- `apps/`: Contains standalone applications
+  - `vscode/`: VS Code extension for Quarto
+  - `writer/`: Web-based Quarto editor
```
cscheid, line 20: I would mention something like "experiment, not in production, ignore this" to avoid Claude going into rabbit holes.

claude.md
```diff
@@ -0,0 +1,35 @@
...
+  - `quarto-core/`: Quarto-specific core functionality
+  - Other utility packages
+
+
+## Additional Resources
+
+- [Quarto Website](https://quarto.org)
+- [Extension on Microsoft marketplace](https://marketplace.visualstudio.com/items?itemName=quarto.quarto)
+- [Extension on Open VSX Registry](https://open-vsx.org/extension/quarto/quarto)
+- [Quarto GitHub Organization](https://github.com/quarto-dev)
```
cscheid, line 35: At the end of the document, I would add

```

# Instructions
```

This is handy when using Claude Code, because the `#` shortcut (to remember things) appends list items to the end of this file, and adding `# instructions` makes those additions readable.

cscheid (commented):
Looks good!

I don't know how much of this information Claude would use, but aside from the comments I made above, none of it seems harmful.

Issue #719, opened 2025-05-23
Positron: Add setting to prefer Positron's bundled Quarto


jmcphers:
Currently, the VS Code/Positron extension treats Positron's bundled copy of Quarto as one of many possible Quartos. Several users have mentioned that they'd like a setting that causes the extension to not only use but _prefer_ this Quarto over others. 

> When run in Positron, I'm finding it useful to set my Quarto path at a project-level to use Positron's bundled Quarto. It forces me to use a recent stable Quarto without me worrying about what I've currently got installed on my PATH (usually dev or prerelease)

This is where we find Positron's Quarto:

https://github.com/quarto-dev/quarto/blob/b45f12ea4f5ec15f1afa916187b9c5eed4098af1/apps/vscode/src/main.ts#L62-L63

The proposed setting would cause this Quarto to take precedence over others, even if they have a later version.

PR #841, opened 2025-10-05
Add setting to use bundled Quarto in Positron

juliasilge:
Addresses https://github.com/quarto-dev/quarto/issues/719

This PR adds a new setting to prefer Positron's bundled Quarto. The setting defaults to false, and I put it between `quarto.path` and `quarto.usePipQuarto` in terms of precedence. It could easily go _below_ `quarto.usePipQuarto`, if we think that's better.

Screenshot 2025-10-05 at 4 11 11 PM

This PR also introduces the use of the `@posit-dev/positron` npm package, as described here:
https://positron.posit.co/extension-development.html
There are a LOT of improvements we could make to the Positron-specific code throughout the extension, but let's do that gradually.

This PR also introduces listening for config changes for all three PATH config options, to prompt the user to reload the window. These settings need a restart to work correctly, or at least for the changes to be reflected in the status bar? But IIUC you really do need a restart.

## How to test this out

- Have a version of Quarto CLI installed that is >1.8
- Open Positron with this new version of the extension (either install from `.vsix` or make a dev build)
- Notice that the status bar tells you that you are using Quarto 1.8+, whatever version you have installed
- Go to settings, find `quarto.useBundledQuartoInPositron`, and enable it
- You should get prompted to reload the window
- Notice that the status bar now tells you that you are using Quarto 1.7.32, the version of Quarto that Positron currently bundles
- Switch the setting back, get prompted to reload, and you are now back to Quarto 1.8+

### DIFF
--- a/apps/vscode/src/main.ts
+++ b/apps/vscode/src/main.ts
@@ -15,6 +15,7 @@
 
 import * as vscode from "vscode";
 import * as path from "path";
+import { tryAcquirePositronApi } from "@posit-dev/positron";
 import { MarkdownEngine } from "./markdown/engine";
 import { kQuartoDocSelector } from "./core/doc";
 import { activateLsp, deactivate as deactivateLsp } from "./lsp/client";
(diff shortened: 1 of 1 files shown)


juliasilge:
I don't think we can add a meaningful test for this because we don't have Positron test runner builds, and we don't have more than one version of Quarto installed in the CLI currently. Those could be areas for future test investment, but probably not the highest priority for testing currently.

cscheid (approved):
I like this (and "what Quarto am I?" is a real source of trouble in our issues, so all help is welcome here). With that said:

- I'm not the right person to do a code review on the positron- and vs-code-specific paths of the code.
- Maybe this is for a separate PR, but it would be good for someone (Positron? the extension?) to surface the result of that quarto-finding logic in a legible way for users, even if only in logs. For example: in the CLI, we have `quarto check`, which produces a report of which versions of dependencies are being used (and where they are in the file system), etc. This way, when something goes wrong, we have a standard flow: "show us `quarto check`, and we'll go from there"

Issue #845, opened 2025-10-07
VSCode opens every file in Visual Editor and breaks formatting


lkhilton:
### Bug description

If I open any file in a workspace in the Visual Editor, from then on any file that I save gets opened in the Visual Editor. This is happening with .qmd files, but also .json, .yaml, .R files, etc. Whenever it does open one of these other files in the Visual editor, it destroys all of the file formatting. 

### Steps to reproduce

- Install the above versions of Quarto and VSCode
- In an empty workspace, create a *.qmd file
- View the file in Visual editor
- Create a *.json file (e.g. `.vscode/settings.json`)
- Save the .json file
- File is opened in Visual editor and .json formatting is collapsed

### Actual behavior

After any .qmd file is opened in the visual editor, all subsequent files are opened in the visual editor with catastrophic effects on file formatting. 

### Expected behavior

The visual editor should only open when requested and only for files of the correct file type (i.e. .qmd files). 

### Your environment

-IDE VS Code 1.96.4
-OS either Mac OS 14.6.1 or CentOS 7 (remote SSH)
-VS Code Quarto extension 1.125.0

Relevant lines from settings.json: 
```
"[quarto]": {
        "editor.formatOnPaste": true,
        "editor.formatOnSave": false,
        "editor.formatOnType": true,
        "editor.wordWrap": "on",
        "editor.quickSuggestions": {
            "comments": "on",
            "other": "on",
            "strings": "on"
        },
        "editor.quickSuggestionsDelay": 250,
        "editor.snippetSuggestions": "none",
        "editor.wordBasedSuggestions": "off",
        "editor.suggestOnTriggerCharacters": true,
        "editor.unicodeHighlight.ambiguousCharacters": false,
        "editor.unicodeHighlight.invisibleCharacters": false
    }
```

Mackey17:
You're not alone. I encountered the same question as you, it was a disaster to my json file.

cscheid:
We're looking into this right now, and will get back to you ASAP.

juliasilge:
I've got a fix here, if @cderv or @cscheid, you can confirm it also works for you: https://github.com/quarto-dev/quarto/pull/848

I can do a release right away if so.

cderv:
Yes this is working.

lkhilton:
Yep - the extension auto-updated this morning and the behavior is back to normal.

PR #848, opened 2025-10-08
Undo `onDidSaveTextDocument` that causes problems for ALL files in VS Code

juliasilge:
Addresses https://github.com/quarto-dev/quarto/issues/845

This PR undoes the change in https://github.com/quarto-dev/quarto/commit/6b416d3e59164cb4bb03e41664e286bc195b83bc that accidentally set up `onDidSaveTextDocument` that applies to _all_ documents. This bug was in a code path that only applied to VS Code.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.126.0 (Unreleased)
 
+- Fixed a bug opening non-Quarto files in visual mode on saving (<https://github.com/quarto-dev/quarto/pull/848>).
+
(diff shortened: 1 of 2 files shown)


juliasilge:
We can confirm the fix by getting the artifact from here: https://github.com/quarto-dev/quarto/actions/runs/18350404038/job/52268505072?pr=848

Download, unzip, then _Extensions: Install from VSIX..._

juliasilge:
You should still be able to switch between visual and source mode in VS Code, but you should _not_ see any attempts to open in visual mode in non-Quarto files on a save.

PR #857, opened 2025-10-22
overview documentation of the extension

vezwork:
Here are some added inline docs, minor refactors, and a large overview markdown file that links to various key places in the code and gives some examples of how parts of the extension communicate with eachother. The inline docs and minor refactors mostly aim to address places I have been getting stuck/lost repeatedly while navigating code over the past few months.

I did this to help me get re-oriented, and to be able to answer questions about how to architect https://github.com/quarto-dev/quarto/pull/825 and future features that involve "piping things through" (sharing functionality between the Source Editor and Visual Editor by communicating from the Visual Editor to the host and LSP in a consistent way).


apps/vscode/src/host/index.ts
```diff
@@ -112,13 +118,20 @@ function defaultExtensionHost(): ExtensionHost {
...
+    registerStatementRangeProvider: doNothing,
+    registerHelpTopicProvider: doNothing,
+    createPreviewPanel: (
+      viewType: string,
+      title: string,
+      preserveFocus?: boolean,
+      options?: WebviewPanelOptions & WebviewOptions
+    ): WebviewPanel => {
+      return window.createWebviewPanel(viewType, title, { viewColumn: ViewColumn.Beside, preserveFocus, }, options);
+    }
```
vezwork: This was previously imported from `preview.ts` where it was defined by itself. I found that hard to navigate and also the implementation of this function in the other ExtensionHost in hooks.ts defines this inline.

apps/vscode/src/providers/cell/commands.ts
```diff
@@ -145,6 +146,8 @@ abstract class RunCommand {
   }
 
   private async hasExecutorForLanguage(language: string, document: TextDocument, engine: MarkdownEngine) {
+    // TODO: this is incorrect right? `cellExecutorForLanguage` returns a promise, and a promise will always be truthy?
+    // We should have to await it before doing `!!`
     return !!this.cellExecutorForLanguage(language, document, engine);
```
vezwork: Didn't want to make a change here because it will change the logic, so I left a TODO so I don't forget the next time I'm working in this area.

apps/vscode/src/host/hooks.ts
```diff
@@ -164,32 +165,18 @@ class EmbeddedStatementRangeProvider implements HostStatementRangeProvider {
...
-          adjustedPosition(vdoc.language, position),
-          vdoc.language
-        );
-      });
+      const result = await vscode.commands.executeCommand<hooks.StatementRange>(
+        "vscode.executeStatementRangeProvider",
+        document.uri,
+        adjustedPosition(vdoc.language, position)
+      );
+      return { range: unadjustedRange(vdoc.language, result.range), code: result.code };
```
juliasilge: This set of changes in this file _breaks_ the statement range provider in source mode. We need to use the code as it was before the change, to propagate the statement range through the virtual doc. We can't call `vscode.executeStatementRangeProvider` on the `.qmd` but instead must call it on the vdoc.

juliasilge (approved):
Looks good! Things here reflect about what I understand to be the facts on the ground here.

Issue #861, opened 2025-11-03
Math formulas render in black in VS Code preview when using dark theme


kv9898:
### Bug description

Summary
Math formulas (KaTeX/LaTeX) in Quarto .qmd preview are rendered in black text regardless of the editor theme. On dark themes that makes inline and display math unreadable.

Reproduction steps

1. Open VS Code and activate a dark theme (e.g., "Dark+ (default dark)").
2. Create a Quarto markdown file (formula.qmd) with a math block, for example:

```qmd
---
title: "Untitled"
format: html
---
# Example formula

This is a formula:

$$
E = mc^2
$$

```

3. Open the Quarto/preview pane (Preview in extension) for the file.
4. Observe the rendered formula text (inline or display math) is black on the dark background.

Actual behavior
Rendered math text is pure black (or a fixed dark color), making it illegible on dark backgrounds.

From the code:

<img width="265" height="124" alt="Image" src="https://github.com/user-attachments/assets/13d8ea34-88b0-4daf-9b28-571fd9f16306" />

From the sidebar:

<img width="242" height="80" alt="Image" src="https://github.com/user-attachments/assets/6141a26d-f8e5-4fa5-a8d2-7a301583947d" />

Expected behavior
Math should inherit the appropriate foreground color from the preview theme or use the editor/preview text color so it remains readable with dark themes. In short: math should respect the page/theme color rather than force black.

Minimal example
Use the YAML/sample above (paste into formula.qmd) — reproduces immediately in preview.

PR #772, opened 2025-07-25
Prototype of getting active vscode theme in the extension

vezwork:
## The problem

As reported in #525, the visual editor does not use code highlighting colors from the user's vscode color theme. 

The extension does get some colors from the theme, but it does not get any code highlighting related colors. I believe the extension includes its own code highlighting color themes for each of the default vscode themes. If the user is using a built-in vscode theme (e.g. Monokai, Solarized, etc.) then the extension is able to guess that and select a matching theme definition from its own list. However, if the user is not using a built-in vscode theme, the extension somehow picks one of the built-in themes to use as the code highlighting them.

You can see this maybe most glaringly with a monochrome theme from the extension store:

| Source Editor with Monochromatic theme   | Visual Editor with same Monochromatic theme |
| -------- | ------- |
| Screenshot 2025-07-25 at 11 24 15 AM | Screenshot 2025-07-25 at 11 23 32 AM  |

---

## Can we get the user's theme colors?

Yes, to some extent at least. This PR so far gives a prototype of getting and logging code highlighting theme data from the user's vscode theme. Here's an example of what we can get:

<details><summary>Here's an example of the data we get</summary>

```
[...]

PR #864, opened 2025-11-05
Fix theme updating when VS Code auto theme detection is on

kv9898:
Fixes https://github.com/quarto-dev/quarto/issues/861

I admit the fix is a bit ugly and I wish better solutions are possible.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Fixed a bug where code blocks inside complex div structures (e.g., many `::: {.notes}` divs without preceding blank lines) were not detected as executable cells (<https://github.com/quarto-dev/quarto/pull/875>).
 - Added a public API that other extensions can query to get the Quarto CLI path, version, and availability (<https://github.com/quarto-dev/quarto/pull/879>).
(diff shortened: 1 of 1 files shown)


vezwork:
Thanks so much for the PR @kv9898 !

I am not working on theme-related stuff right now, but [I will be](https://github.com/quarto-dev/quarto/pull/772), hopefully before the end of the year. I will come back to review your PR then. I will try to help come up with a clean solution to the problem.

kv9898:
@vezwork Happy New Year! Any updates?

vezwork:
Tried this out and it does seem to fix the issue. 

I looked into the solution and it does seem that:
1. using `vscode.window.activeColorTheme.kind` with `onDidChangeActiveColorTheme` is a good way to get the actual current theme kind
2. you cannot directly access the theme kind in the LSP, so you have to send a message from the HOST.
3. We don't currently exactly have a precedent for sending messages from the HOST to the LSP (-- don't have a precedent for sending data, we do send _requests_ in `custom.ts`, maybe we should use that instead?). See: https://github.com/quarto-dev/quarto/blob/326b1e0c68d5bf51234438d9bdf443dc2ae49bb0/apps/overview.md?plain=1#L67-L68
  so it seems reasonable to use the notification messaging functionality of the LSP.

As such, this was pretty close to a minimal solution. I cleaned it up a bit. It is also a necessary fix in other ways because our colorTheme config in the LSP did not accurately represent the current theme previously, it only represented the user's `colorTheme` setting.

Issue #846, opened 2025-10-08
Log Quarto CLI discovery in Quarto output channel


juliasilge:
### Bug description

As @cscheid points out in https://github.com/quarto-dev/quarto/pull/841#pullrequestreview-3314709156, we should surface the result of the logic to find the Quarto CLI in a legible way for users. I think the best option in the "output channel" for the Quarto extension, like here:

<img width="1580" height="1087" alt="Image" src="https://github.com/user-attachments/assets/d4160003-250a-4d65-b22b-aad3a3983163" />

mcanouil:
Is there an exposed API planned to tell third parties that Positron is using bundled version and where it is?

In this case, I'm talking about Quarto Wizard which uses `PATH`, `quarto.path` from the VSCode/Positron extension, but when Quarto CLI is the one bundled, only Positron is able to find it and knows about it, meaning Quarto Wizard cannot use Quarto bundled in Positron currently.

- https://github.com/mcanouil/quarto-wizard/discussions/210

juliasilge:
@mcanouil are you wanting to use it more from the command line within Positron, as described by https://github.com/posit-dev/positron/issues/7605? Or are you wanting to be able to get the path to the bundled Quarto and such from _your_ extension to be able compute something with that path? Or it's awkward that the bundled Quarto is not on the path?

mcanouil:
> Or are you wanting to be able to get the path to the bundled Quarto and such from your extension to be able compute something with that path

This.
Because without, it means Positron users using bundled version of Quarto cannot use Quarto Wizard to install/update extensions, since this requires Quarto CLI.
If the path is exposed by Positron directly or via the quarto extension, then Quarto Wizard could access/use it.

mcanouil:
Thanks for the heads up!

Issue #443, opened 2024-05-15
Quarto extension pollutes Extension Host output channel with a lot of warnings


juliasilge:
In VS Code, the Quarto extension generates quite a lot of warnings that I believe are extraneous in the "Extension Host" output channel:

![Screenshot 2024-05-15 at 9 39 07 AM](https://github.com/quarto-dev/quarto/assets/12505835/358fdb36-9d2d-4f03-a943-a8bd1b22ba50)

For searchability, here is an example:

> [warning] [quarto.quarto] Accessing a resource scoped configuration without providing a resource is not expected. To get the effective value for '[markdown]', provide the URI of a resource or 'null' for any resource.

PR #878, opened 2026-01-02
Fix up some logging (Quarto CLI discovery, extension host warnings)

juliasilge:
This PR addresses:

- the first part of https://github.com/quarto-dev/quarto/issues/846 (logging where/how the Quarto CLI is found)
- https://github.com/quarto-dev/quarto/issues/443 which is soooooo annoying when dealing with the output channels

After these changes, you'll see logs like this in the Quarto output channel, depending of course on the exact details of how your Quarto is configured:

Screenshot 2026-01-02 at 3 41 09 PM

I went fairly verbose on these logs and am open to feedback about this being overkill, but I tend to think more is better, at least to some level. I did not add any of this to the status bar or anything like that; I think that the output channel is a good spot for folks to look for this kind of info.

### DIFF
--- a/apps/vscode/src/core/quarto.ts
+++ b/apps/vscode/src/core/quarto.ts
@@ -60,9 +60,8 @@ export async function configuredQuartoPath(
   const isPositron = tryAcquirePositronApi();
 
   if (useBundledQuarto) {
-    logger?.("  Checking Positron bundled Quarto: enabled");
-
     if (isPositron) {
+      logger?.("  Checking Positron bundled Quarto: enabled");
       // Use path relative to the application root for Positron's bundled Quarto
       const rootPath = env.appRoot; // Use vscode.env.appRoot as the application root path
       const positronQuartoPath = path.join(
--- a/packages/quarto-core/src/context.ts
+++ b/packages/quarto-core/src/context.ts
@@ -30,7 +30,8 @@ export type QuartoSource =
   | "pip-venv"          // pip-installed in venv/conda
   | "path"              // Found on system PATH
   | "known-location"    // Known install location
-  | "additional-path";  // Additional search path (Positron fallback)
+  | "additional-path"   // Additional search path (Positron fallback)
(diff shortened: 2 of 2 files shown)


apps/vscode/src/core/quarto.ts
```diff
@@ -18,29 +18,49 @@ import * as fs from "node:fs";
...
   }
 
   // check if we should use bundled Quarto in Positron
   const useBundledQuarto = config.get("useBundledQuartoInPositron", false); // Default is now false
+  const isPositron = tryAcquirePositronApi();
+
   if (useBundledQuarto) {
-    // Check if we're in Positron
-    const isPositron = tryAcquirePositronApi();
+    logger?.("  Checking Positron bundled Quarto: enabled");
```
vezwork: should this log be inside `if (isPositron)` ?

apps/lsp/src/config.ts
```diff
@@ -158,22 +158,27 @@ export class ConfigurationManager extends Disposable {
...
+        path: quarto?.path ?? this._settings.quarto.path,
         mathjax: {
-          scale: settings.quarto.mathjax.scale,
-          extensions: settings.quarto.mathjax.extensions
+          scale: quarto?.mathjax?.scale ?? this._settings.quarto.mathjax.scale,
+          extensions: quarto?.mathjax?.extensions ?? this._settings.quarto.mathjax.extensions
         },
         symbols: {
-          exportToWorkspace: settings.quarto.symbols.exportToWorkspace
+          exportToWorkspace: quarto?.symbols?.exportToWorkspace ?? this._settings.quarto.symbols.exportToWorkspace
```
vezwork: Nice, more defensive (less error possibilities and risk of things being undefined) and probably more correct.

vezwork (approved):
Approved with questions

vezwork:
Oh! P.s. I like the indentation you've done on the logs!

PR #879, opened 2026-01-03
Add a public API for other extensions to use

juliasilge:
Addresses the last bit of https://github.com/quarto-dev/quarto/issues/846, with a new public API that _this_ extension returns for _other_ extensions to use, to get information about the Quarto CLI being used. This is particularly useful for extensions like Quarto Wizard that need to know the path to the Quarto binary, especially when Quarto is bundled in Positron or installed in a Python virtual environment, etc. 

The new API exposes three methods: 
- `getQuartoPath()`
- `getQuartoVersion()`
- `isQuartoAvailable()` 

Other extensions can access this API via `vscode.extensions.getExtension('quarto.quarto').exports` after copying the interface definition into their own codebase. Since your extension cannot import types from the Quarto extension directly, copy this interface definition into your own codebase:

```typescript
// Copy this interface into your extension
interface QuartoExtensionApi {
  getQuartoPath(): string | undefined;
  getQuartoVersion(): string | undefined;
  isQuartoAvailable(): boolean;
}

// Then use it like this:
async function getQuartoPathFromExtension(): Promise<string | undefined> {
  const quartoExt = vscode.extensions.getExtension('quarto.quarto');
  if (!quartoExt) {
    return undefined;
  }
  if (!quartoExt.isActive) {
    await quartoExt.activate();
  }
  const api = quartoExt.exports as QuartoExtensionApi;
  return api.getQuartoPath();
}
```

This is documented in `api.ts` for now. Eventually we might want to add this documentation elsewhere, but for now, let's see if this works for @mcanouil and then we can evaluate where else we should surface this info.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,7 +2,9 @@
 
 ## 1.128.0 (Unreleased)
 
+- Fixed a bug where code blocks inside complex div structures (e.g., many `::: {.notes}` divs without preceding blank lines) were not detected as executable cells (<https://github.com/quarto-dev/quarto/pull/875>).
 - Added a public API that other extensions can query to get the Quarto CLI path, version, and availability (<https://github.com/quarto-dev/quarto/pull/879>).
+
 
 ## 1.127.0 (Release on 2025-12-17)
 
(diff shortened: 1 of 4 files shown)


mcanouil:
I'll take a look as I'm working on the next release of Quarto Wizard (which turns out to be quite major).

juliasilge:
That's great news! We'll get @vezwork to review this next week and then we can get a new version of the extension released.

vezwork (approved):
Wonderful! Great comments in `apps/vscode/src/api.ts`, thanks so much. Excited about this because it lays some groundwork for exposing internal extension functionality for testing! And you've already set up a test. Great job.

PR #887, opened 2026-01-14
Do not disable Copilot in Quarto files

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/11348

I do not have complete context for why this was originally disabled in 3d697f4e366cfe34fc1fc1bd1f08346882e5911e, but I feel confident that we do not want to disable this now. Before https://github.com/posit-dev/positron/pull/11163, Copilot completions did work for `.qmd` and users have not had problems with the behavior using today's Copilot models.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.129.0 (Unreleased)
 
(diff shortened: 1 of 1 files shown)


cscheid (approved):
Looks good to me! (sending from claude, but manually approved by Carlos)

PR #888, opened 2026-01-16
Fix mistake where changelog entry has added under old release heading

vezwork:
I accidentally added #861's changelog entry under the previous release's (1.128) heading. Correcting that mistake


juliasilge (commented):
We typically don't put this in the URL

PR #895, opened 2026-01-22
Fix bug where visual editor removes the first nested checkbox

### LINKED ISSUE #894: Visual Editor mangles first nested checkbox
### Bug description

Currently, when you load a checklist nested inside a list into the Visual Editor, it removes the first checkbox in the list. For example
```qmd
-

    -   [x] hello
    -   [x] hello
    -   hello
```
loads into the visual editor as:
image
Then saves back (roundtrips) to qmd as:
```qmd
-   [x] 

    -   hello
    -   [x] hello
    -   hello
```

### Steps to reproduce

_No response_

### Actual behavior

_No response_

### Expected behavior

_No response_

### Your environment

_No response_

vezwork:
Fixes #894 

Adds a roundtrip test that passes in this branch but fails on main.

## Cause description

1. the `tokensWithChecked` function called `mapTokens` to detect and modify checkbox nodes in the AST.
2. Checkbox characters were removed in this call so that they don't show up in the Prosemirror document.
3. `mapTokens` acts recursively on the entire sub-AST passed in, so it would modify AST nodes of nested lists, removing their checkboxes.

The bug was extra confusing to diagnose because 1. also meant that `tokensWithChecked` modifies the original AST. I was logging the AST before the call to `tokensWithChecked` to try to understand the issue, but realized after that the logged AST was being modified after being logged, so I was seeing the modified version. 

That's gross. So, in this PR I've also made it so `tokensWithChecked` does not modify the original AST, it instead returns a modified sub-AST.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -8,8 +8,9 @@
 - Fixed cell formatting sometimes deleting code at the end of the cell (<https://github.com/quarto-dev/quarto/pull/754>).
 - Removed visual/source toggle button from Positron editor action bar. You can still switch between editing modes using the command palette or context menu (<https://github.com/quarto-dev/quarto/pull/896>).
 - Improved checkbox list item appearance in the Visual Editor (<https://github.com/quarto-dev/quarto/pull/893>).
-- Fix columns2 and columns3 snippets leaving `:::` in the document (<https://github.com/quarto-dev/quarto/pull/899>).
-- Add fragment snippet (<https://github.com/quarto-dev/quarto/pull/901>).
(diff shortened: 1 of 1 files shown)


packages/editor/src/nodes/list/list-checked.ts
```diff
@@ -212,50 +212,43 @@ export function fragmentWithCheck(schema: Schema, fragment: Fragment, checked: b
...
+    if (modifiedTokens[0].c[1].t === 'Space') {
+      modifiedTokens[0].c.splice(0, 2); // remove the first two tokens
     } else {
-      return tok;
+      modifiedTokens[0].c.splice(0, 1); // remove the first token
     }
-  });
+  }
 
-  // return
```
vezwork: lol

PR #896, opened 2026-01-23
Remove toggle control on Positron action bar for source/visual

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/1805

For now, we are pausing investment in the visual editor in VS Code and Positron, to be able to focus more on the source editor and other alternative editing experiences. This PR removes the prominent toggle button on the Positron editor action bar that was introduced in https://github.com/quarto-dev/quarto/pull/698 so that we don't put this quite so much in user's faces, given that they are likely to have a disappointing experience.

Screenshot 2026-01-23 at 10 57 49 AM

This does not change any of the other ways to get into the visual editor:

- From the command palette
- From the context menu or behind the three dots in the action bar.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,8 +4,9 @@
 
 - Fixed Copilot completions in `.qmd` documents (<https://github.com/quarto-dev/quarto/pull/887>).
 - Fixed a bug where the `autoDetectColorScheme` setting could cause equation previews to have a dark text on dark background and vice versa (<https://github.com/quarto-dev/quarto/pull/864>).
-- Fix a regression where bash cell execution does not work (<https://github.com/quarto-dev/quarto/pull/826>).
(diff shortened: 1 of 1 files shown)


OlalObeja:
> Addresses [posit-dev/positron#1805](https://github.com/posit-dev/positron/issues/1805)
> 
> For now, we are pausing investment in the visual editor in VS Code and Positron, to be able to focus more on the source editor and other alternative editing experiences. This PR removes the prominent toggle button on the Positron editor action bar that was introduced in #698 so that we don't put this quite so much in user's faces, given that they are likely to have a disappointing experience.
> 
> Screenshot 2026-01-23 at 10 57 49 AM
> This does not change any of the other ways to get into the visual editor:
> 
> * From the command palette
> * From the context menu or behind the three dots in the action bar.

Thanks @juliasilge  for this info. Now I understand a few issues better and why some positron visual editor performances were a bit off compared to source mode. Plus also why that familiar source/visual editor toggle is now no longer in my face where it used to be. 

All the way to working in source mode all the time now. Thanks for the work you do!

PR #906, opened 2026-02-04
Add file linking and autocomplete for _quarto.yml file using vscode extension

EmilHvitfeldt:
This PR adds two providers to the Quarto VS Code extension

- `activateYamlLinks`
- `activateYamlFilepathCompletions`

Both can be turned off individually with settings.
Both are set to only trigger if you are in a `_quarto.yml` file as determined by `isQuartoYaml`.

`activateYamlLinks` goes through the document and does some fairly hackish parsing of the the file to find potential file names. Then each match is checked to exist.

`activateYamlFilepathCompletions` triggers on typing in spots that feel right (after `: ` or `- `) then matches what has been typed so far against files from the repo with some hackish ignores `IGNORE_PATTERNS`.

Example gif below:

https://github.com/user-attachments/assets/1ef22c2b-3397-4e07-9175-5eab6dbadaaa

- Am I missing something else for this to work with the extension and quarto in general?
- Do you need tests? I assume so, if yes where
- How do you want this documented for the user? I feel like I typically see features and settings documented in the readme of the extension but you don't do that here

If you provide potential edge cases then I modify the code to account for those. But I wanna make sure i'm on the right track before continuing


vezwork (commented):
I pulled down this PR and built and was not able to get links in a `_quarto.yml` file in VSCode, but I was able to get them in a `_quarto.yaml` file. I noted that syntax highlighting also didn't apply in `_quarto.yml`. I added logs and removed the conditional before this code 
```ts
context.subscriptions.push(
  languages.registerDocumentLinkProvider(
    { language: "yaml", scheme: "file" },
    new QuartoYamlLinkProvider()
  )
);
```
in yaml-links.ts to make sure it was being executed. It seems to be VSCode's file association rather than your code? I haven't used _quarto.yaml/yml files that much, so I'm not sure how to navigate that yet.

Otherwise, the code is looking good (ideally we would be using a yml parser rather than regex, but we aren't really doing that anywhere yet, I think this approach should be fine).

EmilHvitfeldt:
It appears like the setting that you are using, is that all `.yml` files should be treated as quarto files. Which I don't know is a good idea or not. However it make me realize that I was overcomplicating things a bit, and i have updated the code to instead make it so the code runs when you are inside a file named `_quarto.yml` or `_quarto.yaml` instead, which shouldn't depend on user settings

vezwork:
> - Am I missing something else for this to work with the extension and quarto in general?

@cscheid does the file resolution in this PR look reasonable? It would be cool if we had some standard function for resolving files in quarto projects, but I don't think we do, so it seems good to me.

> - Do you need tests? I assume so, if yes where

The closest test we would have so far would be the tests for document symbols in [apps/vscode/src/test/symbols.test.ts](https://github.com/quarto-dev/quarto/blob/main/apps/vscode/src/test/symbols.test.ts). If you were to make tests like these, then you may have to register a vscode command that does something like returning an array of the links present in a yaml file. Such a test would be a little divorced from what the user actually sees in terms of link underlines, but that's the best I can think of. I think up to you if you want to try testing that out, could be good.

> - How do you want this documented for the user? I feel like I typically see features and settings documented in the readme of the extension but you don't do that here

Good question. You should add an entry in the [apps/vscode/CHANGELOG.md](https://github.com/quarto-dev/quarto/blob/main/apps/vscode/CHANGELOG.md). Otherwise, I don't know that we have much of a precedent for this. A good first step would be adding a description of the feature in one or both of the main implementation files I think.

EmilHvitfeldt:
Done!

vezwork (approved):
Great PR. Great comments describing the functionality in each file. Nice looking tests. Please merge whenever

PR #919, opened 2026-02-18
Use `StatementRangeSyntaxError`

DavisVaughan:
Follow up to https://github.com/posit-dev/positron/pull/11907

It feels to me like we should still use the "old" `hooks.d.ts` system for this, because `StatementRange` is already tied up in that `hooks.d.ts` file. Mixing `hooks.d.ts` and types from `@posit-dev/positron` doesn't feel like it would be a very good idea for this extension of an existing feature, IIUC.

We version guard on Positron `2026.03.0` rather than bumping the required version of Positron

Here's a test in the Positron `2026.04.0 build 85` daily build:

https://github.com/user-attachments/assets/3b5312ab-6c51-4738-ad32-e66efd64a8ea

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.131.0 (Unreleased)
 
+- Added support for Positron's statement execution feature that reports the approximate line number of the parse error (<https://github.com/quarto-dev/quarto/pull/919>).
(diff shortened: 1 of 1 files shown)


juliasilge (approved):
This is working great for me! 🙌 

Image

I don't think it's possible to write any meaningful tests for this, without a Positron test runner build.

Can you please update `CHANGELOG.md` for the extension before merging?

PR #946, opened 2026-04-11
feat: improve Quarto snippets

mcanouil:
- Fix a missing comma in the `horizontal rule` prefix array that had rendered the file invalid JSON.
- Fix the raw code block syntax so Pandoc raw attributes use the required `=` prefix (`{=html}` rather than `{html}`).
- Fix `htm` to `html` in the fenced code block language dropdown, drop the non-standard `console`, normalise `c++`/`c#` to `cpp`/`csharp`, and add `bash`, `html`, `r`, `rust`, `sql`, `toml`, `typescript`, `xml`, and `yaml`.
- Add `typst` to the raw code block dropdown.
- Restore the `.column` content lines and add tab stops to `columns2` and `columns3`, drop the trailing `<!-- end columns -->` marker, and append a final `$0` so the cursor exits the block cleanly.
- Normalise final cursor placement with `$0` on single-line snippets (`quote`, `strikethrough`, `heading1`-`heading6`).
- Surround block-level snippets (headings, lists, code blocks, divs, blockquote, horizontal rule) with empty lines so the inserted content does not collide with adjacent paragraphs and, for the horizontal rule, cannot be parsed as a setext heading or YAML block delimiter.
- Add `yaml` for frontmatter, `figure` for an image with `#fig-` id and `fig-alt`, `xref` for a Quarto cross-reference div (`fig-`/`tbl-`/`lst-`), `list-table` for a Quarto `.list-table` div, and `tabset` for a `.panel-tabset` block.
- Update the `div` snippet to use a single placeholder covering id, class, and key-value attributes (`::: {#id .class1 key=value}`) rather than only a class.

### DIFF
--- a/apps/vscode/snippets/quarto.code-snippets
+++ b/apps/vscode/snippets/quarto.code-snippets
@@ -11,7 +11,7 @@
   },
   "Insert quoted text": {
     "prefix": "quote",
-    "body": "> ${1:${TM_SELECTED_TEXT}}",
+    "body": ["", "> ${1:${TM_SELECTED_TEXT}}", "", "$0"],
     "description": "Insert quoted text"
   },
   "Insert inline code": {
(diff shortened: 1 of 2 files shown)


apps/vscode/snippets/quarto.code-snippets
```diff
@@ -108,17 +121,25 @@
...
-    "body": "~~${1:${TM_SELECTED_TEXT}}~~",
+    "body": "~~${1:${TM_SELECTED_TEXT}}~~$0",
     "description": "Insert strikethrough"
   },
   "Insert div block": {
     "prefix": "div",
-    "body": ["::: {.${1:class}}", "${TM_SELECTED_TEXT}$0", ":::"],
+    "body": [
+      "",
+      "::: {${1:#id .class1 key=value}}",
```
vezwork, line 131: Do you you think we should add cursor spots for id, class and key value rather-than/alongside one for all of them?

vezwork (approved):

vezwork:
Tested locally and these changes seem quite nice to have.

apps/vscode/snippets/quarto.code-snippets
vezwork: @mcanouil I'm going to merge this PR now. If you think this is a good idea we can add another small change later.

PR #947, opened 2026-04-14
Terminate Quarto preview when terminal is closed externally

juliasilge:
Addresses posit-dev/positron#13006.

When a user closes the Quarto Preview terminal via the trash icon, the underlying Quarto process was not terminated cleanly. This left behind intermediate `.quarto_ipynb` files, and repeated previews would accumulate `*.quarto_ipynb_1`, `*.quarto_ipynb_2`, etc.

This was happening because VS Code's trash icon sends `SIGHUP` to the shell process only. Quarto, as a child process, typically survives as an orphan and never receives the signal needed to trigger cleanup. <kbd>Ctrl+C</kbd> works because it sends `SIGINT` to the entire foreground process group.

This PR registers an `onDidCloseTerminal` listener in `PreviewManager`. When the preview terminal is closed by any external means, it sends the same HTTP termination request that `killPreview()` already uses, giving Quarto the chance to clean up before it exits. To avoid a duplicate request when the extension itself kills the terminal, `this.terminal_` is cleared before `killTerminal()` is called, so the event handler's identity check short-circuits.


vezwork:
This change looks good to me. Thank you for the clear explanation of the problem and solution. 

I didn't test this PR locally, as I understand it will be verified that the issue was resolved separately.

PR #951, opened 2026-04-16
Use Positron's active runtime when inserting a code cell

### LINKED ISSUE #480: Consult Positron's active language runtime to order code cell types in the Insert Anything tool
It would be good if the VS Code extension would consult Positron's active language runtime to order code cell types in the Insert Anything tool.

This is similar to #425 in terms of improving user experience but suggests consulting one other resource.

juliasilge:
Closes #480

When using _Insert Code Cell_ in source mode, the extension determines the language for the new cell using a priority chain:

1. **Cursor is inside an existing code cell:** use that cell's language and move the cursor to just after the block.
2. **Cursor is outside any code cell, but the document has executable blocks:** use the language of the nearest block (the last one before the cursor, or the first one after if there are none before).
3. **No executable blocks in the document, and running in Positron:** (new, this PR) call `positron.runtime.getForegroundSession()` to get the active runtime's language. If it's one of the supported Quarto languages (`python`, `r`, `julia`, `ojs`, `sql`, `bash`, `mermaid`, `dot`), that language is used.
4. **None of the above:** fall back to an interactive snippet picker listing all supported languages.

Steps 1 to 3 all result in the cursor being placed directly inside the new cell (no picker). Step 4 requires the user to make a selection first.

This is a Positron-only improvement; in VS Code, step 3 is skipped and the picker appears for any empty document.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 - Added clickable document links for file paths in `_quarto.yml` files. File paths are now clickable and navigate directly to the referenced file (<https://github.com/quarto-dev/quarto/pull/906>).
 - Added filepath autocompletion in `_quarto.yml` files. When editing YAML values, the extension now suggests project files as you type (<https://github.com/quarto-dev/quarto/pull/906>).
-- Now use Positron's active runtime to choose the language for new code cells in an empty document (<https://github.com/quarto-dev/quarto/pull/951>).
+- In an empty document, Positron's active runtime is now used to choose the language for a new code cell (<https://github.com/quarto-dev/quarto/pull/951>).
 
 ## 1.131.0 (Release on 2026-04-14)
 
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Added clickable document links for file paths in `_quarto.yml` files. File paths are now clickable and navigate directly to the referenced file (<https://github.com/quarto-dev/quarto/pull/906>).
 - Added filepath autocompletion in `_quarto.yml` files. When editing YAML values, the extension now suggests project files as you type (<https://github.com/quarto-dev/quarto/pull/906>).
+- Now use Positron's active runtime to choose the language for new code cells in an empty document (<https://github.com/quarto-dev/quarto/pull/951>).
 
 ## 1.131.0 (Release on 2026-04-14)
 
--- a/apps/vscode/src/@types/hooks.d.ts
+++ b/apps/vscode/src/@types/hooks.d.ts
@@ -13,6 +13,12 @@ declare module 'positron' {
 		StatementRangeSyntaxError: typeof StatementRangeSyntaxError;
 	}
 
+	export interface LanguageRuntimeSession {
+		readonly runtimeMetadata: {
+			readonly languageId: string;
+		};
+	}
+
 	export interface PositronRuntime {
 		executeCode(
 			languageId: string,
@@ -25,6 +31,8 @@ declare module 'positron' {
 			documentUri: vscode.Uri,
 			cellRanges: vscode.Range[]
 		): Thenable<void>;
+
+		getForegroundSession(): Thenable<LanguageRuntimeSession | undefined>;
 	}
 
 	export interface PositronLanguages {
--- a/apps/vscode/src/providers/insert.ts
+++ b/apps/vscode/src/providers/insert.ts
@@ -116,7 +116,7 @@ class InsertCodeCellCommand implements Command {
         if (language) {
           header = "```{" + language + "}";
         } else {
-          header = "
+          header = "```{${1|" + languages.join(",") + "|}}";
         }
 
         // insert snippet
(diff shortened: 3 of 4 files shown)


apps/vscode/src/@types/hooks.d.ts
```diff
@@ -25,6 +31,8 @@ declare module 'positron' {
 			documentUri: vscode.Uri,
 			cellRanges: vscode.Range[]
 		): Thenable<void>;
+
+		getForegroundSession(): Thenable<LanguageRuntimeSession | undefined>;
```
juliasilge, line 35: I'm adding this here to the `PositronRuntime` interface in along with a minimal `LanguageRuntimeSession` type.

apps/vscode/src/providers/insert.ts
```diff
@@ -98,15 +100,25 @@ class InsertCodeCellCommand implements Command {
           }
         }
 
+        // if no language found in document, fall back to Positron's active runtime
+        const kSupportedLanguages = ['python', 'r', 'julia', 'ojs', 'sql', 'bash', 'mermaid', 'dot'];
+        if (!language) {
+          const api = tryAcquirePositronApi() as unknown as PositronApi | undefined;
```
juliasilge: I'm using `tryAcquirePositronApi()` from `@posit-dev/positron` here to try to get us moved over to the npm package more, but AFAICT I still have to cast to `PositronApi` to get typed access to the new method. I feel like we don't quite have the npm package set up perfectly yet to be used in an extension like this one that is supposed to work in both Positron and VS Code.

juliasilge: I decided this wasn't the right call and changed over in 69f7a7e1d8e666e403c3b81939db903fa6e199d6.

apps/vscode/src/test/insert.test.ts
```diff
@@ -0,0 +1,96 @@
+import * as vscode from "vscode";
+import * as assert from "assert";
+import { examplesOutUri, openAndShowExamplesOutTextDocument, WORKSPACE_PATH } from "./test-utils";
+
+suite("Insert Code Cell", function () {
```
juliasilge, line 5: These tests _only_ test the VS Code behavior, not what would happen in Positron in an empty document. We don't yet have a good way to write tests against Positron APIs.

apps/vscode/src/providers/insert.ts
```diff
@@ -98,15 +99,24 @@ class InsertCodeCellCommand implements Command {
           }
         }
 
+        // if no language found in document, fall back to Positron's active runtime
+        const kSupportedLanguages = ['python', 'r', 'julia', 'ojs', 'sql', 'bash', 'mermaid', 'dot'];
+        if (!language) {
+          const session = await hooksApi()?.runtime.getForegroundSession();
+          const sessionLang = session?.runtimeMetadata.languageId ?? "";
+          if (kSupportedLanguages.includes(sessionLang)) {
```
DavisVaughan, line 107:
```suggestion
        const languages = ['python', 'r', 'julia', 'ojs', 'sql', 'bash', 'mermaid', 'dot'];
        if (!language) {
          const session = await hooksApi()?.runtime.getForegroundSession();
          const sessionLang = session?.runtimeMetadata.languageId ?? "";
          if (languages.includes(sessionLang)) {
```

From what I remember the `kSupportedLanguages` convention is for global variables that live outside functions

apps/vscode/src/providers/insert.ts
```diff
@@ -98,15 +99,24 @@ class InsertCodeCellCommand implements Command {
...
         // if we have a known language, use it and put the cursor directly in the
         // code cell, otherwise let the user select the language first
         let header;
 
         if (language) {
           header = "```{" + language + "}";
         } else {
-          const languages = ['python', 'r', 'julia', 'ojs', 'sql', 'bash', 'mermaid', 'dot'];
-          header = "```{${1|" + languages.join(",") + "|}}";
+          header = "```{${1|" + kSupportedLanguages.join(",") + "|}}";
```
DavisVaughan, line 119:
```suggestion
          header = "```{${1|" + languages.join(",") + "|}}";
```

apps/vscode/CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 - Added clickable document links for file paths in `_quarto.yml` files. File paths are now clickable and navigate directly to the referenced file (<https://github.com/quarto-dev/quarto/pull/906>).
 - Added filepath autocompletion in `_quarto.yml` files. When editing YAML values, the extension now suggests project files as you type (<https://github.com/quarto-dev/quarto/pull/906>).
+- Now use Positron's active runtime to choose the language for new code cells in an empty document (<https://github.com/quarto-dev/quarto/pull/951>).
```
DavisVaughan, line 7:
```suggestion
- In an empty document, Positron's active runtime is now used to choose the language for a new code cell (<https://github.com/quarto-dev/quarto/pull/951>).
```

It sounded odd to me before

DavisVaughan (approved):
Seems like a nice improvement to me

I am indeed annoyed by the fact that it used to choose Python when I had an R Console open _right there_

PR #953, opened 2026-04-22
fix: preserve leading empty lines in code cells (if any) during formatting

### LINKED ISSUE #950: Add empty line after quarto comments
Via @DavisVaughan in https://github.com/posit-dev/air/issues/487:

> Over in https://github.com/jolars/panache/issues/115 we have been discussing an idea like this where the quarto comments are removed before styling, but when they are added back in _an empty line should be added unconditionally_ between the quarto comments and the rest of the code block. That prevents cases like the above from looking so confusing. Can we make this change in Quarto? It seems like it would be an easy tweak.
> 
> So concretely:
> 
> ---
> 
> If we start here
> 
> ```{r}
> #| label: my label
> 
> ## my R comment 1
> a <- 1
> 
> ## my R comment 2
> b <- 2
> ```
> 
> Then we stay there.
> 
> ---
> 
> If we start here
> 
> ```{r}
> #| label: my label
> ## my R comment 1
> a <- 1
> 
> ## my R comment 2
> b <- 2
> ```
> 
> then we add a line and end here
> 
> ```{r}
> #| label: my label
> 
> ## my R comment 1
> a <- 1
> 
> ## my R comment 2
> b <- 2
> ```
> 
> ---
> 
> if we start here
> 
> ```{r}
> #| label: my label
> a <- 1
> b <- 2
> ```
> 
> we add a line and end here
> 
> ```{r}
> #| label: my label
> 
> a <- 1
> b <- 2
> ```
> 
> i.e. that empty line is unconditionally added to split the quarto directives from the rest of the code, even when there isnt a comment in the code itself

mcanouil:
- Formatting an R code cell with Air stripped all leading empty lines between option directives and code. Python cells were unaffected because their virtual documents include inject lines that push user code off position 0.
- `formatBlock` now strips leading empty lines from the virtual document and adjusts the line offset accordingly, so formatter edits land at the correct position.
- A `normalizeEdit` collapses two or more leading empty lines to exactly one; this fires even when no language formatter is active.
- New tests cover: single empty line preserved, multiple collapsed to one, no empty lines unaffected, a hostile formatter proving empty lines are hidden, and normalisation without a formatter — for both R and Python.

Tested interactively on a VSIX installed build with (my formatters: Ruff and Air):

````qmd
---
title: Leading Empty Lines
format: html
---

```{r}
#| label: one-empty-line

x<-1
```

```{r}
#| label: two-empty-lines


x<-2
```

```{r}
#| label: no-empty-lines
x<-3
```

```{python}
#| label: no-empty-line
x=1;y=2
```

```{python}
#| label: one-empty-line

x=1;y=2
```

```{python}
#| label: two-empty-lines


x=3;y=4
```
````

Which lead to:

````markdown
---
title: Leading Empty Lines
format: html
---

```{r}
#| label: one-empty-line

x <- 1
```

```{r}
#| label: two-empty-lines

x <- 2
```

```{r}
#| label: no-empty-lines
x <- 3
```

```{python}
#| label: no-empty-line
x = 1
y = 2
```

```{python}
#| label: one-empty-line

x = 1
y = 2
```

```{python}
#| label: two-empty-lines

x = 3
y = 4
```
````

Fixes #950

### CODE REVIEW

**Comment by cscheid** [comment]
Do I understand it correctly that the code you added here undoes some of the work that Air does?

**Comment by mcanouil** [comment]
Yes😅 (but before it does it, here it does not even know there were empty lines at the top of the code cells)

Ruff/Black would have done the same thing if the extension was not injected a comment at the top of the virtual doc.🤷‍♂️

**Comment by mcanouil** [comment]
To clarify, this PR creates normalisation of the leading empty lines regardless of languages and formatters, this makes things consistent.

**Comment by vezwork** [comment]
@DavisVaughan does this solution seem reasonable to you?

**Review by vezwork — APPROVED** [review]
Tested locally and it works how I would expect. The code and tests look good upon quick review.

File: apps/vscode/src/providers/format.ts [hunk 1]
```diff
@@ -311,14 +312,51 @@ async function formatBlock(
...
+    } else {
+      break;
+    }
+  }
+  const codeLines = blockLines.slice(optionLines + leadingEmptyLines);
+
+  const blockRange = new Range(
+    new Position(block.range.start.line, block.range.start.character),
+    new Position(block.range.end.line, block.range.end.character)
+  );
```
DavisVaughan: Is this different from `block.range`, which already exists?

apps/vscode/src/providers/format.ts
```diff
@@ -311,14 +312,51 @@ async function formatBlock(
...
 
   // Nothing to format if the block is entirely option directives (or only
   // trailing whitespace after them, which `lines()` may produce from a
-  // final newline in `token.data`).
+  // final newline in `token.data`). Still apply normalizeEdit if present.
   if (codeLines.every(l => l.trim() === "")) {
-    return undefined;
+    if (normalizeEdit && !blockRange.contains(normalizeEdit.range)) {
+      return undefined;
+    }
```
DavisVaughan: I don't see how this could happen, it looks like the way you defined `normalizeEdit` would guarantee that it lies within the range of the block

DavisVaughan: (If you remove it here, also remove it again below)

(If you don't remove it, maybe ensure we have a test that shows why it is required)

apps/vscode/src/providers/format.ts
```diff
@@ -351,6 +388,12 @@ async function formatBlock(
     return new TextEdit(range, edit.newText);
   });
 
+  // Include normalizeEdit in the guard so it is validated along with formatter
+  // edits — all edits must be in range or none are applied.
```
DavisVaughan: again, it seems like you built the normalize edit from the original range itself, so this feels circuitous to me, but maybe im missing something.

apps/vscode/src/providers/format.ts
```diff
@@ -311,14 +312,51 @@ async function formatBlock(
...
+    : undefined;
 
   // Nothing to format if the block is entirely option directives (or only
   // trailing whitespace after them, which `lines()` may produce from a
-  // final newline in `token.data`).
+  // final newline in `token.data`). Still apply normalizeEdit if present.
   if (codeLines.every(l => l.trim() === "")) {
-    return undefined;
+    if (normalizeEdit && !blockRange.contains(normalizeEdit.range)) {
+      return undefined;
```
DavisVaughan: Non blocking note: In terms of code style / ease of maintenance, it might be nice if we could structure this function to not have early exists with hard to remember corner cases like the `normalizeEdit` behavior.

So, like, in this case what you might consider doing is assigning `edits = [] / undefined` here when there are no `codeLines` instead of letting `executeFormatDocumentProvider()` run, but then let everything else fall through the "happy path" as much as possible with an empty `edits` array.

In my head, the benefit of this is that the only time `normalizeEdits` would come into play is at the very end, with 

```
  if (normalizeEdit) {
    adjustedEdits.push(normalizeEdit);
  }
```

and otherwise you don't have to think about it, which would be quite nice

apps/vscode/src/test/formatting.test.ts
```diff
@@ -512,6 +525,183 @@ suite("Code Block Formatting", function () {
...
+    );
+    assert.ok(
+      /one-empty-line\n\n/.test(formattedResult),
+      "Single leading empty line should be preserved"
+    );
+    assert.ok(
+      !/one-empty-line\n\n\n/.test(formattedResult),
+      "No extra empty line should be introduced"
+    );
+  });
```
DavisVaughan: Another non blocking note:

I think these kinds of tests are much easier to skim if they are structured as snapshot tests, kind of like https://github.com/quarto-dev/quarto/blob/5f420f8788de39ea4b37f33dd1cba0a3268db5a8/apps/vscode/src/test/vdoc.test.ts

The idea being that for each test you have some on-disk `input.qmd` (here, your `format-r-leading-empty-lines.qmd`) and some known on-disk `output.qmd`. You:
- Format `content(input.qmd)` -> `formatted`
- Compare `content(output.qmd) === formatted`

That way its fairly easy to just look at `output.qmd` and ensure it has all the expected properties we care about

Each test could be structured as a folder, like

```
r-multiple-empty-lines-collapsed-to-one/
  input.qmd
  output.qmd
```

DavisVaughan (approved):

mcanouil:
@DavisVaughan thanks for the review. I'll get back on this PR probably this weekend or next week.

Edit: comments were addressed.

apps/vscode/src/test/formatting.test.ts
vezwork: Also see https://github.com/quarto-dev/quarto/blob/5f420f8788de39ea4b37f33dd1cba0a3268db5a8/apps/vscode/src/test/quartoDoc.test.ts#L71 for another convenient implementation of snapshot testing

apps/vscode/src/providers/format.ts
mcanouil: dbafcf95

apps/vscode/src/providers/format.ts
mcanouil: dbafcf95

apps/vscode/src/providers/format.ts
mcanouil: dbafcf95

apps/vscode/src/providers/format.ts
mcanouil: 49d9c883

apps/vscode/src/test/formatting.test.ts
mcanouil: e9cfc3c3

danielvartan:
Please merge this. It's an OCD nightmare.

cscheid:
> Please merge this. It's an OCD nightmare.

For what it's worth: please don't make this kind of comment in our repos; it's not constructive and only adds to the notification noise we all get.

PR #955, opened 2026-04-26
Commands to convert between `.ipynb` and `.qmd`

### LINKED ISSUE #6: Add `quarto convert` command

seeM:
Closes #6.

This PR adds two commands to convert between `.ipynb` and `.qmd` using the `quarto convert` CLI. The motivation is to ease the onboarding process for Positron users coming from Jupyter Notebooks who are interested in trying out Quarto, and vice versa (https://github.com/posit-dev/positron/issues/11865).

Both commands are available in the command palette and in the UI:

* `quarto.convertToIpynb`: available in source and visual editor overflow menus, under a new group at the bottom
* `quarto.convertToQmd`: available in notebook editor overflow menu (both VS Code notebook editor and Positron notebook editor), under a new group at the bottom

The commands follow the same process:

1. Save the document
2. Check if converting would overwrite a file and prompt the user with a modal dialogue
3. Convert the document
4. Show any tidied up CLI errors
5. Open the converted document

Tests are included.

In future, I plan to add a command to Positron tentatively named `quarto.initializeFromIpynb` that the `quarto.convertToQmd` calls before opening the created `.qmd` file to set the kernel binding and fill the output cache using the original `.ipynb` file for a more seamless transition.


apps/vscode/.vscode-test.mjs
```diff
@@ -2,7 +2,8 @@ import { defineConfig } from '@vscode/test-cli';
 
 export default defineConfig([
   {
-    files: 'test-out/*.test.js',
+    label: 'main',
+    files: 'test-out/!(r-project).test.js',
```
seeM, line 6: This correctly filters out the `r-project` test which has its own configuration below and which currently skips when run from the main configuration.

The `label` is to more easily run _only_ this configuration - before this change it uses the default `--label 0`.

apps/vscode/package.json
```diff
@@ -292,6 +292,16 @@
...
       },
+      {
+        "command": "quarto.convertToIpynb",
+        "title": "Convert Document to .ipynb",
+        "category": "Quarto"
+      },
+      {
+        "command": "quarto.convertToQmd",
+        "title": "Convert Notebook to .qmd",
+        "category": "Notebook"
```
seeM, line 303: Using the `Notebook` category for the `.ipynb` -> `.qmd` command because that's what most notebook commands use, so I expect it to be more discoverable

apps/vscode/package.json
```diff
@@ -722,6 +737,11 @@
         {
           "command": "quarto.previewFormat",
           "group": "q_zPreview"
+        },
+        {
+          "command": "quarto.convertToQmd",
```
seeM, line 747: ~~This should show up correctly in VS Code and in the legacy notebook editor in Positron.~~

~~The Positron notebook editor doesn't connect to the `notebook/toolbar` menu so it'll need additional work to show up there.~~

Nvm, it should work everywhere! I didn't realize that `editor/title` menu items already appear in the Positron notebook editor.

apps/vscode/src/test/r-project.test.ts
```diff
@@ -5,11 +5,6 @@ import * as assert from "assert";
 // projects with a top-level `DESCRIPTION` file.
 
 suite("Workspace Symbols - R Project", function () {
-  // Return early if we're running with the regular test configuration
```
seeM, line 8: No longer needed given the filter change in `.vscode-test.mjs`

apps/vscode/src/providers/convert.ts
```diff
@@ -0,0 +1,234 @@
...
+}
+
+/**
+ * Convert a document using `quarto convert`.
+ *
+ * @param quartoContext Quarto context for running the CLI command
+ * @param outputChannel Output channel for logging errors
+ * @param sourceUri URI of the document to convert
+ * @param targetExt Target extension for the converted file (e.g. ".ipynb", ".qmd")
+ * @returns URI of the converted file if successful, otherwise `undefined`
```
juliasilge: Can we correct the doc comment here, for what this function actually returns?

juliasilge (changes requested):
This is looking great! I have three questions/thoughts:

- Can we improve how error messages get surfaced to the user in terms of escaping? For example, if I have a Quarto file like this:

````
---
title: "unclosed
format: html
---

content

````

And I run the new command, I get this popup, which is not ideal:

Image

- When I convert a `.qmd` to an `.ipynb`, it always opens up the new file in a regular text editor for the JSON, not a notebook editor:

Image

Can we open the new `.ipynb` files in the default notebook editor the user has configured? Either the legacy one or the new Positron notebook editor?

- Can you update the CHANGELOG with info on this new feature?

File: apps/vscode/src/providers/convert.ts (hunk 5 above)
Comment (seeM) [reply apps/vscode/src/providers/convert.ts]: Oops, fixed in 6d7676e

**Comment by seeM** [comment]
Thanks, great catches.

1. The error message now shows only the initial message, and the user can click the Show Output button to see the full error with stack trace (375790c)
    
    image
2. Now uses the `vscode.open` command so it uses the default editor for the file extension, tested with `positron.notebook.enabled` on and off (7a3cd5a)
3. Updated the changelog (b24fdb2)

**Review by juliasilge — APPROVED** [review]
This is looking good and is ready to merge! 🙌 

I want to highlight that if we start sticking a button for this on files in Positron, we are going to come up _hard_ against some of what doesn't work well with `quarto convert`. There are a bunch of situations where it doesn't work well, but here is one example.

Make an R notebook:

Image

Convert and you get this:

````
---
jupyter: python3
---

Here is my great notebook! It will use janitor 🧼🧹

```{python}
library(tidyverse)
library(janitor)

gapminder::gapminder |> 
  clean_names() |> 
  drop_na()
```
````

This isn't a problem with this PR but with `quarto convert` but we will want to be very careful about where we surface this functionality with a button. We want to only put it somewhere that we think it will work well. Maybe no R? And maybe no `.qmd` files, as the conversion from `.qmd` to `.ipynb` is worse than the other way IMO.

seeM:
@juliasilge interesting. What do you think about us fixing those sorts of issues if they come up?

juliasilge:
@seeM Those would need to be fixed in the core of the Quarto CLI, not the extension:

https://github.com/quarto-dev/quarto-cli/issues?q=is%3Aissue%20state%3Aopen%20%22quarto%20convert%22

You are advocating we fix them in the Quarto CLI? 

`quarto convert` is broken enough in enough situations that I don't think we can put a button on, say, a `.qmd` file with R code. Please try it out and see what I mean! I think we'll do better contributing a button in a situation where we think it _will_ work (like an `.ipynb` that uses a Python kernel).

juliasilge:
You can read some more background here: https://github.com/orgs/quarto-dev/discussions/9463

cscheid:
(Just for everyone's awareness, we definitely want to fix these issues; Wasim and I are meeting early next week about this.)

Issue #208, opened 2023-06-26
VScode extension: linting is not working


GitHunter0:
Hey folks,
Linting is not working. 
For example, Jupyter notebooks detect and point the errors below:
![image](https://github.com/quarto-dev/quarto/assets/26294221/7050545b-0cc0-4b25-aa90-29f96a4fd859)
while Quarto does not:
![image](https://github.com/quarto-dev/quarto/assets/26294221/3a13c0ff-7573-4989-9c46-58e84448b7e2)

Also, Quarto HELP Pane does not provide any content in some cases like this one for `glimpse`:
```{python}
import datar.all as d
import pandas as pd
df = pd.DataFrame({'x':[1,2,4]})
d.glimpse(df)
```
Should I open a separate issue for that?

Thank you.

```
Windows 10 pro
python 3.10.4
quarto 1.4.114
Quarto VScode extension v1.88.0
Visual Studio Code 1.79.2
```

jjallaire:
Hi there, linting is known not to work in `.qmd` files (we have issues w/ mapping the Quick Fix positions correctly back into the `.qmd` file). Note that none of the diagnostic engines know how to analyze a `.qmd` so we need to present them with a "fake" e.g. Python script for code intelligence features. In the case of diagnostics the quick fix suggestions don't get mapped correctly back into the `.qmd`.

In terms of Quarto Help, we are there just calling the VS Code "hover" API for the target language --- if you hover over something and see content and then observe that placing the cursor within the same content doesn't drive help then yes file an issue (however if there is no hover, we definitely won't be able to provide help).

GitHunter0:
Thank you for the feedback, @jjallaire .
> linting is known not to work in `.qmd` files
It's a pity, I would love when Quarto allows us to completely replace Jupyter notebooks (which I'm really not a fan).

> if you hover over something and see content and then observe that placing the cursor within the same content doesn't drive help then yes file an issue
Yes, when I hover on `glimpse` in a `.ipynb` file, I get this:
![image](https://github.com/quarto-dev/quarto/assets/26294221/0039d8b0-f89a-4035-90a3-894f6d6689e4)
But when I hover in a `.qmd` file, I get nothing on the Quarto Help pane. I will open a separate issue then.

Bear in mind, that I needed to install [Jupyter Powertoys](https://github.com/microsoft/vscode-jupyter-powertoys) due to this issue https://github.com/microsoft/vscode-jupyter/issues/1307#issuecomment-943658044 .

[...]

PR #957, opened 2026-04-29
Add diagnostics for cells

vezwork:
Addresses https://github.com/quarto-dev/quarto/issues/208

This PR is a prototype of diagnostics for cells. So far it works by creating vdocs in a `.quarto` folder in the same folder of the qmd being edited. This is probably a bad user experience, but LSPs don't seem to want to give diagnostics for vdocs when they aren't in the workspace.

image

So far I've only tested with R code, but the vdocs are prepared in such a way that it should work for python as well.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -970,6 +970,20 @@
           "default": true,
           "markdownDescription": "Show parameter help when editing function calls."
         },
+        "quarto.cells.diagnostics.enabled": {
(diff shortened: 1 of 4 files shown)


juliasilge:
Should we close this, @vezwork, since we took https://github.com/quarto-dev/quarto/pull/980?

vezwork:
Yes, thank you!!

PR #962, opened 2026-05-04
`quarto preview` in a terminal with a venv: increase wait time to 5s

cscheid:
This attempts to work around vscode clobbering the preview command; recent versions appear to inject shell commands after `quarto preview` started, which completely breaks the preview command. The 5s timeout was not determined with any level of rigor, but does consistently solve the problem in my environment.


juliasilge (approved):
This is working well with venv-driven rendering in Positron, with any of the configurations I am using:

Image

I will say that I couldn't reproduce the original problem with VS Code 1.118.1 and the released Quarto extension. Does that surprise you?

Can you add a line to the CHANGELOG for this update?

cscheid:
> I will say that I couldn't reproduce the original problem with VS Code 1.118.1 and the released Quarto extension. Does that surprise you?

I'm not surprised, because this slipped by us. I could only manage to reproduce it by having my shell enable a venv by default.

cscheid:
> Can you add a line to the CHANGELOG for this update?

Sorry, the changelog entry is there now. (My previous push attempt had failed because of the github outage and I didn't notice.)

vezwork:
Added a comment explaining that this timeout is a magic number found with trial & error. Will merge once tests pass

PR #963, opened 2026-05-04
Fix centering of SVG icon

juliasilge:
This PR fixes the centering of the SVG icon:

- Center the `qmd.svg` language file icon within its 16x16 viewBox. The four-quadrant disk was offset ~2 units to the right (centered at x≈10 instead of x=8), making the icon visibly lopsided when rendered.
- All four path coordinates shifted left by 2 units; bounds are now x: 2.5–13.5, y: 2.5–13.5.

Why haven't we noticed this problem?! Because VS Code only renders `contributes.languages.icon` when the user has **no file icon theme active** (File Icon Theme → None). With any active theme, including the default Seti theme or popular ones like Material Icon Theme, VS Code uses the theme's icon instead of the language contribution. Since virtually all users have a file icon theme enabled, the off-center `qmd.svg` was effectively invisible in day-to-day use.

This has come to light because of changes that @dhruvisompura has been making in https://github.com/posit-dev/positron/pull/13162 and related work. It turns out this is extremely noticeable when you change to a different third party file icon theme.


juliasilge:
We'd like to get this merged and released, so we can incorporate these changes into https://github.com/posit-dev/positron/pull/13343.

cscheid (approved):
😱 Looks great now, thank you!

Issue #918, opened 2026-02-17
Feature Request: Add Outline Support for Named Code Sections in R Code Chunks


UrielMenalled:
### Summary

In Positron (but *not* RStudio), named code sections and headers (e.g., `# Section One ----`, `# Section Two ====`) within R code chunks in Quarto documents are recognized and appear in the document outline, aiding navigation. Currently, these sections are not detected in the outline feature when editing Quarto documents, resulting in a loss of navigation usefulness compared to RStudio (see video below).

### Motivation

- Improves Quarto IDE navigation for R users adopting Positron
- Increases parity between RStudio, Positron, and Quarto experience
- Helps users organize complex R scripts within Quarto documents

### Example

A Quarto R code chunk such as:

```{r}
#| label: test

# Section One ----
# Section Two ====
### Section Three ####
```

Should show “Section One”, “Section Two”, and “Section Three” in the outline panel for fast navigation, the way RStudio currently does.

### Related Context

- Used with: R Version 4.4.2, Positron Version 2026.02.0 build 139, macOS 25.2.0 (arm64)
- Reference discussion: https://github.com/posit-dev/positron/discussions/11921

### Feature Request

Please implement recognition of section headers inside R code chunks of Quarto documents, so they appear in the IDE’s Outline/navigation pane, matching RStudio user experience.

https://github.com/user-attachments/assets/45522318-b6d3-44fa-9884-1afe9f1b785c

juliasilge:
This is related to https://github.com/quarto-dev/quarto/issues/880 in that we would need to get the document symbols from the underlying language LSP (so for R or Python, etc) and include them in the document symbols that the Quarto LSP provides.

juliasilge:
Fixed in https://github.com/quarto-dev/quarto/pull/972! 🎉

Issue #880, opened 2026-01-07
Feature Request: In qmd file outline view, in source mode, include R functions (that are in R chunks)


donboyd5:
Outline view for qmd files in source mode shows headers and chunks, but does not show functions within chunks. (This is true for R chunks. I don't know about other languages.)

**Desired feature:** Include functions in the outline.

This is a nice-to-have -- useful in early stages of project development before functions are broken out to separate files, where they do show in outline view and symbol search. It is particularly useful for a long report in qmd/R/typst, because a typst project must be in a single qmd file (or a set of sourced qmd files) rather than a book project, so the qmd file can be long and include many functions until they are consolidated in R files.

**Fallback desired feature:** Include functions in qmd file symbol finder (e.g., ctrl+t on my Windows computer, apparently cmd+t on some computers). They don't currently show.

Apologies if this is discussed already. I looked but did not find it.

Note: Initially posted in [Positron #11237](https://github.com/posit-dev/positron/issues/11237).

Positron Version: 2026.01.0 (user setup) build 138
Code - OSS Version: 1.106.0
Commit: https://github.com/posit-dev/positron/commit/703dfcdfd72914c5a57ab5843d546433995e52d9
Date: 2025-12-24T07:49:32.328Z
Electron: 37.7.0
Chromium: 138.0.7204.251
Node.js: 22.20.0
V8: 13.8.258.32-electron.0
OS: Windows_NT x64 10.0.26200

juliasilge:
This would involve getting document symbols from the underlying language LSP (so for R or Python, etc) and including them in the document symbols that the Quarto LSP provides.

juliasilge:
Fixed in https://github.com/quarto-dev/quarto/pull/972! 🎉

Issue #167, opened 2023-05-07
Cell names do not show in document outline (VSCode)


nickriches:
Cell names are really useful for navigating. By this I mean

```{python CELL_NAME}
CODE
```

This works well in RStudio to help navigate to relevant code cells/chunks. However, I have not found a way to do this in Quarto when using VSCode. When I view the outline the names of code cells do not appear.

jjallaire:
Thanks for this, we are hoping to make some big improvements in outlining/indexing over the next few months and will make code blocks part of this.

nickriches:
Excellent - thanks! 😊

promicrobial:
This would be a super nice feature to have! I've been making do with creating subheadings for different code chunks.

renatovargas:
Thank you for this. I'm coming from posit-dev/positron#3822 and was instructed to bring it up here instead. 

Not an issue, but a shameless request related to this. Great work btw. For Quarto documents (.qmd), RStudio shows only headings and subheadings in the outline, like you can see below on the left. Positron, on the other hand, shows headings, subheadings **_and_** (code cell) like you can see on the right. Now this looks great for defined code chunk listings that are named for cross refs, but very crowded for the rest of unnamed code chunks.

![image](https://github.com/user-attachments/assets/31e7ed87-6c08-4265-9c75-59d72be54a83) ![image](https://github.com/user-attachments/assets/9d723409-6fa2-4db5-a5f8-15fc6b983986)

I am aware that you can click on "collapse all" like below, but that will collapse to first level headings only and you lose some ability to quickly navigate the document, especially in this case, where you go back and forth, perhaps trying out different elasticities in a very specific subsection to see poverty results at the end or going back to see how you set something up. I am aware also that you can individually open up each heading.

![image](https://github.com/user-attachments/assets/e1e02803-0db8-47c4-85dc-337807323046)

Now, I'm not saying to get rid of "(code cell)", because it encourages you to name your chunks, but I would appreciate it if there was perhaps a "toggle code cells" option, independent from collapse/expand for the Outline panel, so that you could limit it to headings and subheadings in Rmd, qmd's and other document type files. Thank you.

artmg:
Returning to the main issue of code cells appearing in the Outline and breadcrumbs as `(code cell)`, I just wanted to observe that the VScode feature 'Sticky Header on Scroll' **does** correctly capture code cell boundaries and their name (even if it makes no attempt to parse the name from the line) as shown here:
[...]

PR #972, opened 2026-05-06
Add code symbols into outline

vezwork:
Addresses an issue mentioned by @juliasilge: https://github.com/quarto-dev/quarto/issues/167#issuecomment-4083541698 

Screenshot 2026-05-06 at 4 04 39 PM

Adds code symbols under code cells in the outline.


vezwork:
I attempted to add tests until I remembered that our test setup does not have actual LSPs for Python or R in it, so we can't really test this. Well... now that I write that, I realize I could probably mock one out. That seems a little contrived, but maybe still worth it.

juliasilge:
This PR:

- fixes https://github.com/quarto-dev/quarto/issues/880
- fixes https://github.com/quarto-dev/quarto/issues/918

juliasilge (commented):
I seem to be getting two of everything, at least in Positron:

Image

vezwork:
I was able to reproduce duplicated language symbols in code cells in Positron with inline outputs enabled. Looking into what is happening.

vezwork:
@juliasilge In `qmd`s in Positron with `Positron › Quarto › Inline Output: Enabled` enabled, the following command (from `getCodeCellSymbols` in this PR) is returning duplicated symbols:
``` ts
await commands.executeCommand<DocumentSymbol[] | SymbolInformation[]>(
  "vscode.executeDocumentSymbolProvider",
  uri
);
```
I checked the surrounding context, including the vdoc content. Everything seems to be in order, nothing strange otherwise. When I disabled `Positron › Quarto › Inline Output: Enabled` and re-open the document, I no longer experience the duplicated symbols issue. It appears to me that `the "vscode.executeDocumentSymbolProvider"` command does not function as expected when inline outputs are enabled.

vezwork:
@seeM mocked out fake LSPs in https://github.com/quarto-dev/quarto/pull/980. Maybe I can copy that approach for tests in this PR.

vezwork:
Cell Diagnostics tests were failing with a message about timing out in CI. The current run of the tests passed, but there wasn't a code change that would've affected the tests or the code tested. The tests seem to be flakey for some reason. Let's keep our eye on that.

apps/vscode/src/lsp/client.ts
```diff
@@ -364,6 +372,120 @@ function isWithinYamlComment(doc: TextDocument, pos: Position) {
...
+    if (!vdoc) return undefined;
+
+    // Get symbols from the embedded language server
+    return await withVirtualDocUri(vdoc, document.uri, "completion", async (uri: Uri) => {
+      try {
+        const result = await commands.executeCommand<DocumentSymbol[] | SymbolInformation[]>(
+          "vscode.executeDocumentSymbolProvider",
+          uri
+        );
+        if (result.length === 0) return undefined;
```
juliasilge: The `result` here could also be nothing at all (`undefined` probably?) so it might be nicer to handle that in this `if` so it doesn't show up as an error in that `catch()`.

apps/vscode/src/lsp/client.ts
```diff
@@ -364,6 +372,120 @@ function isWithinYamlComment(doc: TextDocument, pos: Position) {
...
+
+        if (isDocumentSymbol(result[0])) {
+          return unadjustSymbolRanges(result as DocumentSymbol[], vdoc.language, cellRange.start.line);
```
juliasilge: Is it possible to also handle results that come back as `SymbolInformation[]`, as indicated in line 457?

juliasilge (commented):
I looked into the issue with inline output and it _is_ some change we'll need to make over on the Positron side; I'll log that once this is merged.

I do think we need to add at least some basic tests, with a fake in-process symbol provider (not a real LSP).  We already do this for the formatting tests. I think the steps would be:

1. Register a temporary provider in `symbols.test.ts` with `vscode.languages.registerDocumentSymbolProvider({ scheme: "file", pattern: "**/.vdoc.*" }, ...)` so it only handles Quarto virtual docs.
2. Make that provider return each shape in separate tests:
   - `DocumentSymbol[]`
   - `SymbolInformation[]`
   - `undefined`
3. Run `vscode.executeDocumentSymbolProvider` on `format/basics.qmd` or another example file and assert:
   - no throw for `undefined`
   - code-cell symbols still exist
   - returned embedded symbol names appear under/alongside code-cell outline entries for both result shapes.

That will give us coverage of the middleware logic without having to set up real LSPs.

apps/vscode/src/lsp/client.ts
vezwork: How do you know it could be nothing at all? The https://code.visualstudio.com/api/references/commands seems to say that it will always resolve to a list (although it is slightly ambiguous if the type of the list is `(SymbolInformation | DocumentSymbol)[]` or if its `SymbolInformation[] | DocumentSymbol[]`). Google search summary says it resolves to `SymbolInformation[] | DocumentSymbol[]` but with no good citation. Do you know a way to figure out the return type of a command? I attempted to find the definition of the command in the vscode repo, but couldn't.

apps/vscode/src/lsp/client.ts
vezwork: yes, committed a change for this.

apps/vscode/src/lsp/client.ts
vezwork: In my testing I have now come across it being `undefined`!

vezwork:
Ok I added tests but frustratingly was not able to get them to work when run in sequence, only in isolation. Doubly confusing is that mocking a SymbolInformation provider somehow results in duplicated results in the outline in the test, but idk why. Couldn't figure it out today :|

juliasilge:
In a0506c9488f1304597afa0c6508a0425730f0aa6 I made some changes to how we run this new testing. The `SymbolInformation[]` test was skipped because it failed when run after other tests but passed in isolation, and showed duplicated symbols. The cause was not `provider.dispose()` actually in the end. What I believe it was instead is that `vscode.executeDocumentSymbolProvider` caches results per document URI (VS Code's `OutlineModel` cache). Because every test opened the same `format/basics.qmd`, a stale cached symbol tree from a previous test (or another suite that opened the same file) was served instead of re-invoking the middleware. That leaked the previous test's symbols, dropped the current test's, and surfaced duplicates.

I added a new helper function to address this in `test-utils`, which opens a unique on-disk copy of an example file so each test gets a fresh URI and the provider actually runs. I only set it up to be used in the new code cell symbol tests but this likely will be useful for other LSP features that work like this with caching.

juliasilge:
I opened https://github.com/posit-dev/positron/issues/13907 to fix the duplicated symbols with inline output.

PR #973, opened 2026-05-07
Add custom pair colorization and highlighting for divs

### LINKED ISSUE #670: Allow folding and colorisation of fenced div
- https://code.visualstudio.com/api/language-extensions/language-configuration-guide

Current "workaround" at least for the paired colorisation:

```json
"editor.guides.bracketPairs": "active",
"editor.language.brackets": [
  ["[", "]"],
  ["(", ")"],
  ["{", "}"],
  ["::::", ":::"]
],
"editor.language.colorizedBracketPairs": [
  ["[", "]"],
  ["(", ")"],
  ["{", "}"],
  ["::::", ":::"]
],
```

vezwork:
Fixes #670 (except "Navigation"? What does that mean? Do we want divs in the outline?)

Screenshot 2026-05-07 at 3 47 08 PM

Adds colorization to pairs of opening and closing div syntax. It is a custom decorator. I don't believe we can use VSCode's builtin pair colorizer because it relies on opening and closing brackets being preset and distinct, whereas Quarto fenced divs can have any number of colons >2 in them and the closing and opening brackets are the same (other than the trailing class/attribute syntax `{}` on the opening bracket).

Also fixes the parsing of opening syntax for divs where the class/attribute syntax is empty i.e. exactly `::: {}`. This was previously not parsed as opening a div (it was parsed as an inline), the regex expected at least one character inside `{}`. As a result, folding also works better now.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.133.0
 
-- Add custom pair colorization and highlighting for divs in qmds (<https://github.com/quarto-dev/quarto/pull/973>).
+- Added custom pair colorization and highlighting for divs in qmds (<https://github.com/quarto-dev/quarto/pull/973>).
 
 
 ## 1.132.0 (Release on 2026-05-05)
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## 1.133.0
 
+- Add custom pair colorization and highlighting for divs in qmds (<https://github.com/quarto-dev/quarto/pull/973>).
+
+
 ## 1.132.0 (Release on 2026-05-05)
 
 - Added clickable document links for file paths in `_quarto.yml` files. File paths are now clickable and navigate directly to the referenced file (<https://github.com/quarto-dev/quarto/pull/906>).
--- a/apps/vscode/src/providers/div-brackets.ts
+++ b/apps/vscode/src/providers/div-brackets.ts
@@ -1,7 +1,7 @@
 /*
  * div-brackets.ts
  *
- * Copyright (C) 2025 by Posit Software, PBC
+ * Copyright (C) 2026 by Posit Software, PBC
  *
  * Unless you have received this program directly from Posit Software pursuant
  * to the terms of a commercial license agreement with Posit Software, then
@@ -15,6 +15,7 @@
 
 import * as vscode from 'vscode';
 import { markdownitParser, Token } from 'quarto-core';
+import { createThrottle } from '../core/throttle';
 
 /**
  * Provides colored decorations for div bracket pairs (:::)
@@ -25,11 +26,18 @@ import { markdownitParser, Token } from 'quarto-core';
 export function activateDivBracketDecorations(context: vscode.ExtensionContext) {
   const parser = markdownitParser();
 
+  // Read debounce delay from config
+  const getDelayMs = () =>
+    vscode.workspace.getConfiguration('quarto').get('cells.background.delay', 250);
+
   // Cache for parsed tokens
   const parseCache = new Map<string, {
     version: number;
     divTokens: Token[];
   }>();
+
+  // Map of editors to their throttled update functions
+  const editorThrottledFunctions = new Map<vscode.TextEditor, () => void>();
 
   // Define decoration types for different nesting levels (rotating colors)
   const decorationTypes = [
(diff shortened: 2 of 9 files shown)


apps/vscode/src/providers/div-brackets.ts
```diff
@@ -0,0 +1,173 @@
+/*
+ * div-brackets.ts
+ *
+ * Copyright (C) 2025 by Posit Software, PBC
```
juliasilge:
```suggestion
 * Copyright (C) 2026 by Posit Software, PBC
```

apps/vscode/src/providers/div-brackets.ts
```diff
@@ -0,0 +1,173 @@
...
+    vscode.window.onDidChangeActiveTextEditor(editor => {
+      if (editor) {
+        triggerUpdateDecorations(editor);
+      }
+    })
+  );
+
+  // Update decorations when document changes
+  context.subscriptions.push(
+    vscode.workspace.onDidChangeTextDocument(event => {
```
juliasilge, line 139: Should we debounce this the same way we do the highlighting decorations in `background.ts`?

apps/vscode/src/providers/div-brackets.ts
```diff
@@ -0,0 +1,173 @@
...
+ * Provides colored decorations for div bracket pairs (:::)
+ *
+ * This gives visual feedback similar to bracket pair colorization,
+ * but works for Quarto's context-sensitive div syntax.
+ */
+export function activateDivBracketDecorations(context: vscode.ExtensionContext) {
+  const parser = markdownitParser();
+
+  // Cache for parsed tokens
+  const parseCache = new Map<string, {
```
juliasilge, line 56: Could we add another listener in here so that we delete this cache at `onDidCloseTextDocument`?

juliasilge (commented):
This is looking great!

Image

Can we add some basic tests for this, in case we find a change we need to make later? I would think something like a new `divs.test.ts` that is similar to `codeBlocks.test.ts`, opening a `.qmd` and confirming the parser finds a `Div` token. Pretty cheap to test through `MarkdownEngine`! If you are motivated to go a little further, you could extract the logic for the depth calculation and `getDivMarkerRange` regex from `activateDivBracketDecorations` into a pure function to unit test, but that might not be worth it.

apps/vscode/src/providers/div-brackets.ts
vezwork: I went down a bit of a rabbit hole. 

- Yes we should not be calling `updateDecorations` too often because that could cause lag.
- BUT! background.ts's debounce was implemented incorrectly -- a new debounced function was getting created on every call of the trigger function (we have to call the same debounced function multiple times for the debounce to take effect).
- Also BUT! I think throttling should be a nicer user experience than debouncing. With throttling, the editor will update as the user is typing, but at a reduced rate, and it will definitely have an update based on the user's last input.

so I made both div-brackets.ts and background.ts use throttling for their updates.

apps/vscode/src/providers/div-brackets.ts
vezwork: yes, done

apps/vscode/src/providers/div-brackets.ts
vezwork: Also we can't (easily) use lodash's throttle/debounce because it has a fixed delay, whereas we want the delay to be from the config.

vezwork: I'm reconsidering my decision to use throttle instead of debounce. I am consulting the #frontend channel in Slack to see if anyone has any thoughts.

vezwork: Thanks to @dhruvisompura and @dotNomad for very helpful discussion and resources on Slack around throttling v.s. debouncing.

I tested a 32000 line qmd and did not experience any input lag when creating and removing divs with throttled bracket highlighting. That only says so much, because my work laptop is powerful, but at least there isn't an egregious perf issue here.

In summary, I think throttling is a better user experience than debouncing here!

apps/vscode/CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 ## 1.133.0
 
+- Add custom pair colorization and highlighting for divs in qmds (<https://github.com/quarto-dev/quarto/pull/973>).
```
juliasilge:
```suggestion
- Added custom pair colorization and highlighting for divs in qmds (<https://github.com/quarto-dev/quarto/pull/973>).
```

Just for consistency in this file

juliasilge (approved):
Amazing! 🤩 This is looking fantastic:

Image

bryce-carson:
@vezwork divs in the outline is a definite feature request; h1 through h6 and div often serve the same purpose, but some longer form collapsible details in divs deserve an outline entry whereas they don't otherwise change the table of contents.

PR #974, opened 2026-05-08
Add setting and command to show/hide cells in outline

### LINKED ISSUE #647: Toggle to show/hide code cells in Outline
Creating a separate issue to kindly request if it's possible to implement a toggle option to show/hide code cells in Outline, as suggested by @renatovargas here:

> Thank you for this. I'm coming from posit-dev/positron#3822 and was instructed to bring it up here instead. 
> 
> 
> 
> Not an issue, but a shameless request related to this. Great work btw. For Quarto documents (.qmd), RStudio shows only headings and subheadings in the outline, like you can see below on the left. Positron (and now I know VSCode), on the other hand, shows headings, subheadings **_and_** (code cell) like you can see on the right. Now this looks great for defined code chunk listings that are named for cross refs, but very crowded for the rest of unnamed code chunks.
> 
> 
> 
> image image
> 
> 
> 
> I am aware that you can click on "collapse all" like below, but that will collapse to first level headings only and you lose some ability to quickly navigate the document, especially in this case, where you go back and forth, perhaps trying out different elasticities in a very specific subsection to see poverty results at the end or going back to see how you set something up. I am aware also that you can individually open up each heading.
> 
> 
> 
> image
> 
> 
> 
> Now, I'm not saying to get rid of "(code cell)", because it encourages you to name your chunks and it's a superb feature, but I would appreciate it if there was perhaps a "toggle code cells" option, independent from collapse/expand for the Outline panel, so that you could limit it to headings and subheadings in Rmd, qmd's and other document type files. Thank you.
> 
>  

 _Originally posted by @renatovargas in [#167](https://github.com/quarto-dev/quarto/issues/167#issuecomment-2246062082)_

vezwork:
Fixes #647

https://github.com/user-attachments/assets/d148f2dd-d561-4204-9ba9-f7b04525e28b

- Adds a `symbols.showCodeCellsInOutline` setting to the `Quarto` configuration,
- Adds a `quarto.toggleCodeCellsInOutline` command to toggle that setting.

Note: I attempted to add tests but for some reason in the tests the outline / document symbols refuse to change. I tried many things including waiting longer during the test after calling `quarto.toggleCodeCellsInOutline`, programmatically getting document symbols with `vscode.executeDocumentSymbolProvider`, looking with my own eyes at the outline during the test. I'm not sure why it doesn't work in the test environment!

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.133.0
 
+- Add setting and command to show/hide cells in outline (<https://github.com/quarto-dev/quarto/pull/974>)
+
 ## 1.132.0 (Release on 2026-05-05)
 
 - Added clickable document links for file paths in `_quarto.yml` files. File paths are now clickable and navigate directly to the referenced file (<https://github.com/quarto-dev/quarto/pull/906>).
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## 1.133.0
 
-- Add setting and command to show/hide cells in outline (<https://github.com/quarto-dev/quarto/pull/974>)
+- Added setting and command to show/hide cells in outline (<https://github.com/quarto-dev/quarto/pull/974>)
 
 ## 1.132.0 (Release on 2026-05-05)
 
--- a/apps/vscode/src/providers/symbols.ts
+++ b/apps/vscode/src/providers/symbols.ts
@@ -24,7 +24,8 @@ class ToggleCodeCellsInOutlineCommand implements Command {
     const currentValue = config.get<boolean>("symbols.showCodeCellsInOutline", true);
     const newValue = !currentValue;
 
-    // Update the configuration - the `registerOutlineConfigListener`config listener handles outline refresh
+    // The LSP re-registers its document symbol provider on config change, which triggers VS Code to re-query the outline.
+    // The VS Code extension restores outline expansion state after the re-query (see `registerOutlineConfigListener`).
     await config.update("symbols.showCodeCellsInOutline", newValue, vscode.ConfigurationTarget.Global);
 
     vscode.window.showInformationMessage(
--- a/apps/vscode/src/providers/symbols.ts
+++ b/apps/vscode/src/providers/symbols.ts
@@ -37,3 +37,53 @@ class ToggleCodeCellsInOutlineCommand implements Command {
 export function symbolsCommands(): Command[] {
   return [new ToggleCodeCellsInOutlineCommand()];
 }
+
+
+
+const expandOutline = async (uri: vscode.Uri) => {
+  // make sure document can provide symbols (for the outline) before expanding the outline
+  await vscode.commands.executeCommand("vscode.executeDocumentSymbolProvider", uri);
+  await vscode.commands.executeCommand("outline.expand");
+};
+/**
+ * Executes `listener(editor)` ONCE, the next time the user switches their active text editor to a qmd.
+ */
+const onNextChangeActiveTextEditorToQmd = (listener: (editor: vscode.TextEditor) => any) => {
+  const listenForNextChangeToQmdDisposable =
+    vscode.window.onDidChangeActiveTextEditor((editor) => {
+      if (editor?.document.languageId === "quarto") {
+        // once we switch to a quarto file once, stop listening
+        listenForNextChangeToQmdDisposable.dispose();
+        listener(editor);
+      }
+    });
+};
+
+/**
+ * Restore outline expansion state after settings that affect symbol output change.
+ *
+ * The LSP re-registers its document symbol provider whenever the relevant
+ * settings change, which forces VS Code to re-query and refresh the outline.
+ * That re-query rebuilds the tree from scratch, so VS Code's heuristic for
+ * symbols with newly-appearing children defaults them to collapsed (e.g.
+ * toggling on code cells leaves their parent headers collapsed).
+ *
+ * We expand the outline once a Quarto editor is active: immediately if the
+ * user already has one focused (e.g. they ran the toggle command), or on the
+ * next switch back if the setting was changed from the Settings UI.
+ */
+export function registerOutlineConfigListener(context: vscode.ExtensionContext) {
+  context.subscriptions.push(
+    vscode.workspace.onDidChangeConfiguration((event) => {
+      if (event.affectsConfiguration("quarto.symbols.showCodeCellsInOutline")) {
+        if (vscode.window.activeTextEditor?.document.languageId === "quarto") {
+          expandOutline(vscode.window.activeTextEditor.document.uri);
+        } else {
+          onNextChangeActiveTextEditorToQmd((editor) => {
+            expandOutline(editor.document.uri);
+          });
+        }
+      }
+    })
+  );
+}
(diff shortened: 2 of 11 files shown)


apps/vscode/src/providers/symbols.ts
```diff
@@ -0,0 +1,44 @@
+/*
+ * symbols.ts
+ *
+ * Copyright (C) 2025 by Posit Software, PBC
```
juliasilge:
```suggestion
 * Copyright (C) 2026 by Posit Software, PBC
```

juliasilge (commented):
This setting right now is only read when we start the app, so the experience is that the user changes the setting but doesn't observe the behavior change. I think we need to either:

- add a listener for config change for this setting, OR
- add a note to the setting description like "Restart to apply" or similar

vezwork:
> This setting right now is only read when we start the app, so the experience is that the user changes the setting but doesn't observe the behavior change.

In my testing in both VSCode and Positron, when I use the "Quarto: Toggle Code Cells in Outline" command the change in the outline is instantly reflected. Are you not experiencing that in your testing? Or do you mean if the user changes the setting manually we should notify them to restart?

juliasilge:
> Or do you mean if the user changes the setting manually we should notify them to restart?

Yep, that's right. Right now, if the user changes the setting, nothing happens until the next time they close/open the app or reload the window. I think the best choice is to use `vscode.workspace.onDidChangeConfiguration` and a listener like `if (event.affectsConfiguration("quarto.symbols.showCodeCellsInOutline"))`. A less good option that would still be acceptable is to put a note on the setting description that the user will need to restart VS Code/Positron to see the change in behavior.

It also would be worth re-examining the no-op edit hack and seeing if we can either:

- have the client send a `workspace/didChangeConfiguration` to the LSP to retrigger the outline generation, OR
- in the config listener, call `vscode.commands.executeCommand("vscode.executeDocumentSymbolProvider", uri)` on the active doc, so the toggle command just flips the setting, and the listener owns the refresh

vezwork:
@juliasilge I tried both the methods you mentioned and more, but was not able to get the outline to refresh without using the no-op edit hack (VSCode seems to have an outline cache that we can't bust without making an edit to the document?). 

I've pulled out the no-op edit hack into `onDidChangeConfiguration` with `if (event.affectsConfiguration("quarto.symbols.showCodeCellsInOutline"))` like you suggested, so that we try to refresh the outline no matter how the config was changed.

I tested and this method of refreshing the outline seems to work well.

https://github.com/user-attachments/assets/78fbedd4-d278-42e5-aaa2-ad61b8385141

@juliasilge what do you think? Too hacky and we should tell the user to refresh or make an edit themselves? It seems to work well, so I think I would lean towards using the hack.

juliasilge:
@vezwork I pushed 338c1459dcd9730bb375302b357288b6891b60c2 on top of your work that takes a different swing at the no-op edit hack, plus fixes a related cold-start bug that has been contributing to bad behavior in this area. We can revert that commit if we don't think this is a good way to go but I'd love to pitch you on this different approach! 😅 

Part of the bad experience around the outline and why you were needing that no-op edit hack is that reloading the window (or opening VS Code with a previously-open `.qmd` restored) leaves the outline empty until you close and reopen the file. This was due to the same root cause as why your hack was needed in the first place:

`apps/lsp/src/index.ts` registers `onDocumentSymbol` in `onInitialize`, but `mdLs` isn't created until partway through `onInitialized`, after several `await`s. Between those events, VS Code sends `textDocument/documentSymbol`, the handler returns `mdLs?.getDocumentSymbols(...) || []` (empty), and VS Code caches that empty response. Without a document change to bust the cache, the outline stays empty until you close/reopen.

There are four main things to keep track of in my commit there:

1. **Dynamic registration of the document symbol provider** in `onInitialized` once `mdLs` exists, dropping the static `documentSymbolProvider: true`. Matches the pattern `diagnostics.ts` already uses. The capability simply isn't visible to VS Code until we can serve it, so the cold-start problem goes away.

2. **Re-register on `configManager.onDidChangeConfiguration`** to bust the client's symbol cache when the setting changes. The LSP spec has no `workspace/documentSymbol/refresh`, but re-registration is the spec-shaped equivalent. No more synthetic text edit firing `onDidChangeTextDocument` to every extension watching.

3. **`outline.expand` after the re-query settles**, in a new `registerOutlineConfigListener` in `apps/vscode/src/main.ts`. Re-registration causes VS Code to rebuild the tree collapsed (its heuristic when symbol shape changes), and `outline.expand` restores the visual state. The listener also handles the Settings UI path, too; if you change the setting while a non-Quarto tab is focused like the Settings tab, the expand waits until you switch back to a Quarto editor.

4. **`mdLsReady` promise** that `onDocumentSymbol` and `onFoldingRanges` await. Not strictly required for symbols anymore (dynamic registration handles it), but folding ranges has the same race so the gate is useful there too.

There are some tradeoffs compared to what you had in this PR. Your no-op edit hack approach preserved outline expansion state natively; mine loses it on re-registration and we have to explicitly restore it with `outline.expand`. I do think avoiding the no-op edit is better because the synthetic edit fires `onDidChangeTextDocument` to every extension watching the doc, and re-registration is the LSP-spec way to invalidate a symbol cache. There is a little remaining quirk still; when the user toggles "show cells", headers that were previously collapsed get expanded too. Probably fine since toggling is rare and most people doing it want to see the cells.

Take a look and see what you think! I'd love to get a better, less risky approach here.

apps/vscode/CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1.133.0
 
+- Add setting and command to show/hide cells in outline (<https://github.com/quarto-dev/quarto/pull/974>)
```
juliasilge:
```suggestion
- Added setting and command to show/hide cells in outline (<https://github.com/quarto-dev/quarto/pull/974>)
```

For consistency with the rest of this file

juliasilge:
I was also able to add tests for this in dc5b7cb95683f2bdf2dae21700f6e96f3a5a89f0.

vezwork:
@juliasilge I am completely sold on your pitch. Not only is your approach less hacky, it also feels nicer to use because the outline refreshes more immediately.

I refactored the code a lil bit:
- instead of waiting a magic 200ms before expanding the outline, instead we `await vscode.commands.executeCommand("vscode.executeDocumentSymbolProvider", uri);` to ensure document symbols are provided first. I think this should more reliably ensure the outline is populated before expanding it? In my testing it works.
- moved the outline expansion listener to symbols.ts.
- refactored away `pendingExpand` and the ongoing active text editor listener in favour of adding a active text editor change listener when the config changes, then expanding the outline and removing the listener once the user changes their active text editor to a qmd. 
- removed some `await`s that did not have any effect.

vezwork:
My one (maybe nitpick) concern is that if the user has multiple qmds open, only the first qmd they visit after changing their `show/hide outline cells` config will have an expanded outline. I tested and it is mildly annoying. I can't think of how to address that without expanding the outline of every qmd the user ever switches to. Maybe its okay to ignore this concern for now?

vezwork:
Overall I'm happy with where this PR is at now and would be down to merge if you are @juliasilge .

vezwork:
Thanks for the tests btw! Did you have any trouble getting them working?? 

(edit: maybe tests work now because your approach is less hacky!?)

juliasilge (approved):
> Did you have any trouble getting them working??

Not really; I think the problem before was all the bad timing with `onDocumentSymbol` but once that was resolved in a better way, the tests could "just work".

Thank you for the nice improvements in 1c6bda38e0b93aa26bd83009ddd7fecc6c5950ad!

PR #979, opened 2026-05-15
Update `yarn dev-vscode` to build tests with `dev` mode (i.e. watch and recompile)

seeM:
Previously, if you ran the VS Code extension's tests from the test pane it would not trigger a recompile and would run stale tests. You had to use `yarn test-vscode`.

This PR updates `yarn dev-vscode` (more specifically, `apps/vscode/build.ts`) to also build tests with `dev` mode (i.e. a watch server that automatically recompiles).

### DIFF
--- a/apps/vscode/build.ts
+++ b/apps/vscode/build.ts
@@ -26,6 +26,7 @@ const testBuildOptions = {
   outdir: 'test-out',
   external: ['vscode', 'mocha', 'glob'],
   sourcemap: true,
+  dev,
 };
 
 const defaultBuildOptions = {
@@ -36,4 +37,11 @@ const defaultBuildOptions = {
   dev
 };
 
-runBuild(test ? testBuildOptions : defaultBuildOptions);
+if (test) {
+  runBuild(testBuildOptions);
+} else if (dev) {
+  runBuild(defaultBuildOptions);
+  runBuild(testBuildOptions);
+} else {
+  runBuild(defaultBuildOptions);
+}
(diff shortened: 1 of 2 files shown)


vezwork:
@seeM I am trying to understand the workflow you are describing. What is the test pane? Is this it? I don't see any tests here when I open the repo.
image

My personal workflow is to keep `yarn dev-vscode` running in the background while I'm working in the repo, I think this continuously recompiles the tests right? But if I understand correctly, it does not feel good to be able to run stale tests (regardless of workflow) with the `yarn test-vscode` command. This PR makes it so running `yarn test-vscode` first recompiles right? Does it also keep the build water running in the background after?

vezwork:
@seeM explained to me in Slack:

> Today:
> - `yarn dev-vscode` does not compile tests
> - `yarn test-vscode` does recompile (on every run) and run tests
>
> Running a test from the extension above does not recompile tests
> 
> After PR:
> - `yarn dev-vscode` additionally watches & compiles tests
> - `yarn test-vscode` as before
> 
> Running a test from the extension above still does not recompile, but dev-vscode will now handle that for us

He is using this extension to populate the test pane with extension tests: https://marketplace.visualstudio.com/items?itemName=ms-vscode.extension-test-runner

vezwork:
I tried this PR locally and was still getting stale tests when using the extension. I noticed the `test-out` dir was not getting cleared, and after manually deleting it and rebuilding, the stale tests went away. I added a commit that adds a small custom plugin to the build that clears the outdir (`test-out` in the case of the test build) before putting the built files back into it.

vezwork (approved):
Working well for me locally! A bit hard to test, but it should address all stale test issues we've been facing and doesn't have much of a downside.

PR #980, opened 2026-05-15
Cell diagnostics

### LINKED ISSUE #208: VScode extension: linting is not working

seeM:
Fixes #208, joint work with @vezwork, building on #957. Here's how it works as described in the original PR:

> This PR is a prototype of diagnostics for cells. So far it works by creating vdocs in a .quarto folder in the same folder of the qmd being edited. This is probably a bad user experience, but LSPs don't seem to want to give diagnostics for vdocs when they aren't in the workspace.

We improved on that by creating vdocs in the temp dir when possible (pyrefly and ark) to keep it out of the user's workspace, and by deleting the vdoc once diagnostics are received.

The vscode-R extension is an exception and requires vdocs in the workspace, so we put it under `.quarto`.

Here's the full sequence diagram:

```mermaid
sequenceDiagram
    actor User
    participant VSC as VS Code
    participant Mgr as EmbeddedDiagnosticsManager<br/>(Quarto extension)
    participant LC as LanguageClient<br/>(vscode-languageclient)
    participant LS as Language Server<br/>(e.g. Pyrefly, Ark)

    User->>Mgr: User action initiates diagnostic request<br/>e.g. document open or edit (after debounce)
    activate Mgr
    Mgr->>Mgr: Parse and group cells by language

    loop For each language found in the document
    create participant VDoc as Virtual Document<br/>(temp file on disk)
    Mgr->>VDoc: Create temp file with code<br/>per language
    Mgr->>VSC: openTextDocument(vdoc URI)
    deactivate Mgr
    activate VSC
    VSC->>LC: onDidOpenTextDocument(vdoc)
    deactivate VSC
    activate LC
    LC->>LS: textDocument/didOpen (vdoc)
    deactivate LC
    activate LS
    LS-->>LS: Analyze document

    LS->>LC: textDocument/publishDiagnostics
    deactivate LS
    activate LC
    LC->>VSC: Set diagnostics for vdoc URI
    deactivate LC
    activate VSC
    VSC-->>User: ⚠️ Vdoc entry briefly visible<br/>in Problems pane (known issue)
    VSC->>Mgr: onDidChangeDiagnostics(vdoc URI)
    deactivate VSC
    activate Mgr
    par Publish .qmd cell diagnostics
    Mgr->>Mgr: Filter diagnostics to code cell ranges
    Mgr->>VSC: Set diagnostics for .qmd URI
    activate VSC
    VSC->>User: Squiggly underlines in editor +<br/>entries in Problems pane
    deactivate VSC
    and Cleanup
    Mgr->>VSC: Set vdoc language to plaintext
    activate VSC
    destroy VDoc
    Mgr-xVDoc: Delete temp file
    deactivate Mgr
    VSC->>LC: onDidCloseTextDocument(vdoc)
    deactivate VSC
    activate LC
    LC->>LS: textDocument/didClose
    deactivate LC
    activate LS
    LS->>LC: textDocument/publishDiagnostics (empty)
    deactivate LS
    activate LC
    LC->>VSC: Clear diagnostics for vdoc URI
    deactivate LC
    activate VSC
    VSC-->>User: Vdoc entry removed from Problems pane
    deactivate VSC
    end
    end
```

### Clearing vdoc diagnostics

This also includes a fix to get vdoc diagnostics to _actually clear_. The problem is that VS Code constrains extensions' ability to close text documents and fire `onDidCloseTextDocument`, which language clients forward to language servers as `textDocument/didClose` notifications, and which they generally respond to by clearing diagnostics.

One option was to use `WorkspaceEdit` to delete the file which _would_ fire the event, but it doesn't let us skip the trash.

The hack in this PR is to change the vdoc's language (set it to `plaintext`), which sends `textDocument/didClose` and clears vdoc diagnostics.

### Known issues

1. There can still be a momentary flicker in the Problems pane of a vdoc's diagnostics. We could probably filter those out in Positron core, but don't think there's anything short-term we can do for VS Code.
2. This won't work for R in Positron just yet. We need a tiny PR to Positron to update the R language client middleware to stop hiding vdoc diagnostics.

### Tests

There is an extensive e2e test suite included. It uses a fake language server, but otherwise exercises the full production code paths, and covers a bunch of edge cases that I ran into while working on this.

There's a bunch of test infra added that might be useful for testing the other language features in future.


seeM:
@vezwork @juliasilge I'm not sure who would be the best person to review this. It is big (although ~half is tests and test infra) so I'm happy to walk either/both of you through it in a call.

DavisVaughan:
> We improved on that by creating vdocs in the temp dir when possible (pyrefly and ark) to keep it out of the user's workspace

If a user opens `foo.qmd` of:

```
my-workspace/
  foo.qmd
  bar.R
```

One eventual (but rapidly coming) goal of ark is that diagnostics on `foo.qmd` should be relative to other files in the workspace, like `bar.R`

But if you were to open, say, `~/Desktop/baz.qmd` while in the `my-workspace` workspace, then diagnostics for `baz.qmd` should not "learn" anything from `bar.R` because `baz.qmd` is not "under" the workspace.

So if you create `tmp/virtual-foo.R` in a temp dir, then that's going to look roughly equivalent to the `~/Desktop/baz.qmd` case, right? It's going to miss valuable context from `bar.R` and other scripts that live in the workspace.

That may not be exactly how ark works today, but we are rapidly iterating our way towards this kind of design

apps/vscode/src/test/test-utils.ts
```diff
@@ -83,3 +83,18 @@ ${RESET_COLOR_ESCAPE_CODE}`);
...
+ * Races a promise against a timeout, returning `undefined` if
+ * the timeout is reached before the promise resolves.
+ */
+export async function raceTimeout<T>(promise: Promise<T>, ms: number): Promise<T | undefined> {
+  let timeout: NodeJS.Timeout;
+  const timeoutPromise = new Promise<undefined>((resolve) => {
+    timeout = setTimeout(() => resolve(undefined), ms);
+  });
+  return Promise.race([
+    promise.finally(() => clearTimeout(timeout)),
```
vezwork, line 97: nit: can just race against `wait(ms)` instead? I don't think clearing the timeout is necessary.

apps/vscode/src/vdoc/languages.ts
```diff
@@ -23,6 +23,11 @@ export interface EmbeddedLanguage {
   emptyLine?: string;
   comment?: string;
   trigger?: string[];
+  /**
+   * Lines of code to inject at the top of the virtual document.
+   * Used to disable diagnostics for virtual documents that were
```
vezwork:
```suggestion
   * Used to add lines that hint to language LSPs to disable diagnostics for virtual documents that were
```

apps/vscode/src/vdoc/languages.ts
```diff
@@ -88,6 +93,11 @@ interface LanguageOptions {
   type?: "content" | "tempfile";
   localTempFile?: boolean;
   emptyLine?: string;
+  /**
+   * Lines of code to inject at the top of the virtual document.
+   * Used to disable diagnostics for virtual documents that were
```
vezwork:
```suggestion
   * Used to add lines that hint to language LSPs to disable diagnostics for virtual documents that were
```

apps/vscode/src/vdoc/vdoc-tempfile.ts
```diff
@@ -21,44 +21,37 @@ import * as uuid from "uuid";
...
   Position,
   TextDocument,
   Uri,
   workspace,
-  WorkspaceEdit,
 } from "vscode";
 import { VirtualDoc, VirtualDocUri } from "./vdoc";
 
+interface VirtualDocTempFileOptions {
+  /** Fire a hover request to prime the language server before returning. */
```
vezwork:
```suggestion
  /** Fire a "dummy" hover request to cause the language server to start  */
```

apps/vscode/build.ts
vezwork, line 47: Didn't we [merge these changes](https://github.com/quarto-dev/quarto/pull/979/changes) into [`main`](https://github.com/quarto-dev/quarto/blob/main/apps/vscode/build.ts) already? Not sure why these are showing up as changes in this PR.

vezwork (approved):

vezwork:
Tested locally and I am getting red and blue underlines in R and Python cells in VSCode, and red and blue underlines in Python cells in Positron.

vezwork:
rebased on main, updated changelog, and force pushed (excuse me, didn't mean to add myself as author to all these commits :|)

apps/vscode/src/providers/diagnostics.ts
```diff
@@ -0,0 +1,579 @@
...
+      `[EmbeddedDiagnostics] Received ${rawDiagnostics.length} diagnostics for ` +
+      `${session.language.ids[0]} in ${workspace.asRelativePath(session.documentUri)}`
+    );
+
+    // Filter: only keep diagnostics that map to a real language block.
+    const mapped: Diagnostic[] = [];
+    for (const diagnostic of rawDiagnostics) {
+      const block = languageBlockAtPosition(session.languageBlocks, diagnostic.range.start);
+      if (block !== undefined) {
+        mapped.push(new Diagnostic(diagnostic.range, diagnostic.message, diagnostic.severity));
```
juliasilge: Is it not possible map the whole `Diagnostic` here, with the other components that may be offered by the LSP? You can read more here:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic

apps/vscode/src/providers/diagnostics.ts
```diff
@@ -0,0 +1,579 @@
...
+      await Promise.all(docSessions.map(s => this.activateSession(s, document, tokens)));
+    }
+  }
+
+  private async recreateSessionsForDocument(document: TextDocument): Promise<void> {
+    await this.removeSessionsForDocument(document.uri, "document-changed");
+    await this.createSessionsForDocument(document);
+  }
+
+  private async activateSession(
```
juliasilge, line 298: IIUC this awaits the vdoc being created and we could get a race between real document changes and a stale vdoc. Can we keep old async work from mutating manager state after a newer session has taken over? Something in here where we check if it's still the current session?

juliasilge (commented):
This is working remarkably well as I put it through its paces:

Image

I am really impressed. 😅 

I will say that in the long run for Positron, I am still going to advocate that we take a different approach. I think an in-memory representation of `.qmd` that is very, very notebook-like will be a better option, both for LSP features as well as AI features. I am actually so eager for us to do that work as outlined in https://github.com/posit-dev/positron/issues/12837 that I would advocate we do that _instead_ of making the Ark LSP work with this new approach in this PR. That new document model approach also sets us up for the LSP work that @DavisVaughan mentions that would be truly difficult/impossible with `vdoc` files.

That being said, it is incredible to me that this works and I think we should merge it, pending some of the minor feedback on the PR. This would be a really huge win for VS Code (i.e. not Positron) users.

apps/vscode/src/test/test-utils.ts
seeM: Good question. I wasn't sure so looked into it and it turns out that if you don't `clearTimeout`, the Node process won't exit until the timeout elapses. Here's a repro:

```sh
node --input-type=module -e "
  let s = Date.now();
  process.on('exit', () => console.log((Date.now() - s) + 'ms'));
  await Promise.race([
    Promise.resolve(),
    new Promise(r => setTimeout(r, 5000))
  ]);
"
```

apps/vscode/build.ts
seeM: Thanks for the rebase! The only remaining change is to add the new test `fixtures` and `utils` folders which is new

seeM:
@DavisVaughan with this approach we can create the vdoc temp files next to the `.qmd` if we'd prefer better diagnostics over cleaner user workspace. Longer-term, the alternate approaches we've chatted about would give us both.

apps/vscode/src/providers/diagnostics.ts
seeM: Good idea! Done in 299255b. Now we can see the language server name, diagnostic code, etc in `.qmd` files:

Screenshot 2026-05-26 at 15 43 53

apps/vscode/src/providers/diagnostics.ts
seeM: Ah nice catch. Fixed in 2f438bf.

seeM:
@vezwork @juliasilge I fixed a few more edge cases and CI is passing reliably (reran a few times and only failures were in other tests). Could you please take another look?

juliasilge (approved):
This continues to work well for me in Positron, for both R and Python:

https://github.com/user-attachments/assets/f3573bb4-a293-44cc-84fc-acd5ef99dbe8

I'm excited to see how this approach plays out for users and definitely think we should merge/release this, although like I said before, I am also interested in us pursuing an option that avoids `vdoc` files on disk altogether, for Positron.

PR #986, opened 2026-05-29
Bump Mermaid to 11.12.0, to match CLI

### LINKED ISSUE #13881: Upgrade Mermaid to New Version
#### Positron and OS details:

Positron Version: 2026.05.2 (user setup) build 3
Code - OSS Version: 1.109.2
Commit: 1b5dfba9314510340cb08f2fedccc5ffba822beb
Date: 2026-05-13T19:16:42.268Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26200

#### Session details:

R 4.6.0.

## Describe the issue:

Mermaid that ships with Positron is outdated at version 9.3.0. 
This means the new shapes introduced at 11.3.0. cannot be rendered and we get "undefined". 

## Steps to reproduce the issue:

Quarto document
1. Add snippet like

```{mermaid}
flowchart TD
    A@{ shape: text, label: "This is a text block" }
```

like the new syntax at https://mermaid.ai/open-source/syntax/flowchart.html#terminal-point-stadium

Preview display says "undefined"

## Expected or desired behavior:

Preview displays diagrams correctly.

juliasilge:
Fixes https://github.com/posit-dev/positron/issues/13881

This PR bumps the Mermaid version used by the Diagram preview webview from **9.3.0** to **11.12.0**, matching the version already bundled in the Quarto CLI.

## How far behind were we?

Soooooooo far! 😱 The Diagram preview was pinned to Mermaid **9.3.0** (late 2022), while the Quarto CLI ships **11.12.0**. That is **two full major versions** behind, plus dozens of minor releases and roughly two and a half years of drift. So a diagram could render correctly in `quarto render` / `quarto preview` (CLI) and then break in the in-editor Diagram preview, purely because the preview was running ancient Mermaid.

The user-visible symptom that surfaced this is that the newer node-shape syntax added in Mermaid 11.3.0, e.g.

```
flowchart TD
    A@{ shape: text, label: "This is a text block" }
```

renders as "undefined" in the preview, because 9.3.0 has no idea what that syntax means.

## What changed?

- **`assets/www/diagram/mermaid.min.js`**: replaced the bundled 9.3.0 build with 11.12.0, copied verbatim from the CLI source (`quarto-cli/src/resources/formats/html/mermaid/mermaid.min.js`) so the two stay in lockstep. The CLI build appends `globalThis["mermaid"] = ...`, which keeps `window.mermaid` available to the webview's classic `<script>` load, so no CSP or script-loading changes were needed.
- **`assets/www/diagram/diagram.js`**: migrated to the Mermaid v10+ async API.
  - `await mermaid.parse(src)` (parse is now async)
  - `const { svg } = await mermaid.render(id, src)` then inject the returned markup, replacing the removed callback-style `render(id, src, cb)`
  - dropped the deprecated `mermaid.mermaidAPI` accessor in favor of `mermaid.initialize(...)`
  - the webview `message` handler now awaits the render so `render-begin` / `render-end` still bracket the actual render and a queued state update cannot interrupt an in-flight one
  - error reporting falls back through `err.str || err.message || String(err)` since v11 errors do not always carry `.str`

Here is what we see afterward:

Screenshot 2026-05-29 at 2 58 11 PM

It's not a perfect match for what you see if you do `quarto preview` or `quarto render`; if more folks use this preview we might want to improve it.

FWIW The OJS stdlib pin (`packages/ojs/external-observablehq-stdlib/src/dependencies.js`, Mermaid 9.2.2) is a separate code path and is intentionally left untouched here because I don't feel very confident about it.

It might be worth thinking about updating this more regularly. 😅


juliasilge:
I added an extension-level test in ebb071a628fb41d51c6bae7197e517926110b8eb that will warn us the next time we get out of sync on this.

apps/vscode/src/test/mermaidVersion.test.ts
```diff
@@ -0,0 +1,131 @@
...
+
+/**
+ * Emit a GitHub Actions warning annotation (surfaced on the PR and in the run
+ * summary). Outside Actions it just prints a line, which is harmless.
+ * See https://docs.github.com/actions/using-workflows/workflow-commands-for-github-actions
+ */
+function emitActionsWarning(title: string, message: string): void {
+  const data = message.replace(/%/g, "%25").replace(/\r/g, "%0D").replace(/\n/g, "%0A");
+  console.log(`::warning title=${title}::${data}`);
+}
```
vezwork: Let's pull this out to the utils file?

vezwork (approved):
Having trouble pulling down this to test locally while travelling. The code looks good to me. The description is clear, with one question: why do you say the mermaid preview pane doesn't match Quarto render? Is it an obvious difference or something we will have to get nuanced about?

juliasilge:
> why do you say the mermaid preview pane doesn't match Quarto render?

It's mostly about the styling. When you do `quarto render` it uses one styling:

Screenshot 2026-06-01 at 6 22 25 PM


When you use this little preview functionality, it uses a different styling:

Screenshot 2026-06-01 at 6 23 14 PM


I don't think we need to optimize for that right now but could keep our ears open for feedback about it.

PR #989, opened 2026-05-30
fix: enable Julia cell execution in Positron

### LINKED ISSUE #988: Support 'Julia for Positron' extension (ntluong.julia-positron) for Julia cell execution in Positron IDE
## Summary

When using Quarto `.qmd` files with `engine: julia` in **Positron** (the data science IDE by Posit, built on VS Code), executing Julia code chunks fails with:

> Executing julia cells requires the Julia extension.

## Root Cause

The Quarto VS Code extension checks specifically for the `julialang.language-julia` extension to enable Julia cell execution. However, `julialang.language-julia` is **incompatible with Positron** due to a VS Code engine version mismatch:

```
Cannot install 'julialang.language-julia' extension because of version mismatch in Marketplace
```

Positron currently ships with VS Code engine `1.109.2`, while the latest `julialang.language-julia` requires a higher engine version.

## The Fix

There is a community extension — **Julia for Positron** (`ntluong.julia-positron`) — specifically built to provide Julia REPL/console integration within Positron's architecture. Quarto should recognize this extension as a valid Julia provider in addition to `julialang.language-julia`.

## Request

Please update the extension detection logic to also accept `ntluong.julia-positron` (or any extension that exposes the required `executeInREPL` API) so that Julia cell execution works in Positron without requiring the incompatible `julialang.language-julia`.

## Environment

- IDE: Positron 1.109.2
- Extension: Julia for Positron (`ntluong.julia-positron`)
- OS: macOS

DevenVGokhale:
## Summary

- Adds `case "julia"` to the Positron hooks executor (`hooks.ts`) so Julia chunks in `.qmd` files route through Positron's runtime API (`workbench.action.positronConsole.executeCode`), the same way Python and R already do
- Adds `ntluong95.positron-julia` as an accepted Julia extension ID in `executors.ts`, since `julialang.language-julia` is incompatible with Positron due to a VS Code engine version mismatch

## Root cause

In `hooks.ts`, the `cellExecutorForLanguage` switch only handled `"python"`, `"csharp"`, and `"r"`. Julia had no case, so the function returned `undefined`, causing:

```
Cannot read properties of undefined (reading 'executeInREPL')
```

The fix is a one-line addition of `case "julia":` to fall through into the same Positron runtime executor path used by R and Python.

Additionally, `julialang.language-julia` cannot be installed in Positron (VS Code engine version mismatch), so the community extension `ntluong95.positron-julia` is the only available Julia extension for Positron users. Adding it to `requiredExtension` allows the extension check to pass.

## Test plan

- [ ] Open a `.qmd` file with `engine: julia` in Positron with `ntluong95.positron-julia` installed
- [ ] Confirm executing a `{julia}` code chunk sends code to the Positron Julia console
- [ ] Confirm variables appear in the Variables pane and plots in the Plots pane
- [ ] Confirm Python and R chunk execution is unaffected

Closes #988

🤖 Generated with [Claude Code](https://claude.com/claude-code)


juliasilge (approved):
Thank you so much for this contribution, @DevenVGokhale! We appreciate it so much 🙌 

Working great, at least at this level:

Screenshot 2026-06-27 at 2 30 41 PM

PR #1001, opened 2026-06-11
fix: enable preview commands in any ipynb editor

### LINKED ISSUE #9792: Positron Notebooks: Preview/render with Quarto
Provide actions in the notebook toolbar to preview/render a notebook with Quarto, similar to what we have in .qmd files:

Image

Also the Preview Format command in the additional/secondary actions.

Image

seeM:
Enables "Preview" and "Preview Format" commands in any `.ipynb` editor, since they do work in that case. Fixes https://github.com/posit-dev/positron/issues/9792.

Here's what it looks like in the Positron Notebook Editor:

image

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 1.134.0 (Unreleased)
 
+- The "Preview" and "Preview Format..." commands now show in the Positron Notebook Editor overflow menu (<https://github.com/quarto-dev/quarto/pull/1001>).
+
 ## 1.133.0 (Release on 2026-06-03)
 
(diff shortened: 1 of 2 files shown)


vezwork (approved):
Haven't tested yet, but if you've tried it in both Vscode and Positron, and after you add a changelog entry, please go ahead and merge (if you have permissions to). Otherwise I'll test, add changelog, and merge later.

vezwork:
I tested this on main after it was merged. I downloaded the ipynb from [https://jupyter.org/try-jupyter/lab/](https://jupyter.org/try-jupyter/lab/) and opened it in the notebook editor in vscode. I then clicked the three dots in the upper right and clicked Preview. I got an info toast "No Quarto document available to render" and nothing else happened. I got the same result when I clicked Preview Format. Notably, convert to .qmd does work fine.

PR #1002, opened 2026-06-11
fix: render in positron notebook

seeM:
Enable "Render Document" in all .ipynb editor tabs. Related to https://github.com/posit-dev/positron/issues/9792.

Here's what it looks like in the Positron Notebook Editor:

image


juliasilge (approved):
I honestly feel bad when I have to explain the difference between preview and render to people. 😭 

However, thank you!

seeM:
@juliasilge yeah, it is confusing. We have some related issues in the Positron repo scheduled for next month, might be a good opportunity to think about if that can be improved

PR #1003, opened 2026-06-12
Fix notebook preview

seeM:


apps/vscode/src/core/doc.ts
```diff
@@ -58,18 +53,19 @@ function isLanguageDoc(languageId: string, doc?: vscode.TextDocument) {
...
-export function isNotebook(doc?: vscode.TextDocument) {
-  return !!doc && isNotebookUri(doc.uri);
+function isNotebook(doc?: vscode.TextDocument | vscode.NotebookDocument): doc is vscode.NotebookDocument {
+  return !!doc && 'notebookType' in doc;
 }
 
-export function isNotebookUri(uri: Uri) {
+function isIpynbUri(uri: Uri) {
   return extname(uri.fsPath).toLowerCase() === ".ipynb";
 }
```
vezwork, line 62: 1. Nit: prefer `dog !== undefined` over `!!doc` for messiness of truthiness reasons

2. Could you add a comment explaining why `'notebookType' in doc` is the thing to check for `isNotebook`? Is notebook that same as ipynb? If we're changing `isNotebookUri` to `isIpynbUri`, should we also change `isNotebook` to `isIpynb`?

apps/vscode/src/providers/preview/commands.ts
```diff
@@ -56,17 +56,15 @@ abstract class PreviewDocumentCommandBase extends RenderCommand {
...
         !hasRenderOnSave ||
         (hasRenderOnSave && format) ||
         !(await isPreviewRunningForDoc(targetEditor.document));
       if (render) {
         if (format === kChooseFormat) {
 
-          const frontMatter = targetEditor.notebook
-            ? targetEditor.notebook.cellAt(0)?.document.getText() || ""
-            : documentFrontMatterYaml(this.engine_, targetEditor.document);
+          const frontMatter = targetEditor.frontMatterYaml(this.engine_);
```
vezwork, line 67: why is there less cases than before? It doesn't look like `targetEditor.frontMatterYaml`  in `editor.ts` captures all these cases?

seeM:
I decided to take a less drastic approach for now, see https://github.com/quarto-dev/quarto/pull/1007.

PR #1007, opened 2026-06-15
fix preview in notebooks

### LINKED ISSUE #1006: Render/Preview fail in notebooks that are visible but not active or when Panel/Sidebar are visible
> I tested this on main after it was merged. I downloaded the ipynb from [https://jupyter.org/try-jupyter/lab/](https://jupyter.org/try-jupyter/lab/) and opened it in the notebook editor in vscode. I then clicked the three dots in the upper right and clicked Preview. I got an info toast "No Quarto document available to render" and nothing else happened. I got the same result when I clicked Preview Format. Notably, convert to .qmd does work fine.

_Originally posted by @vezwork in https://github.com/quarto-dev/quarto/issues/1001#issuecomment-4684755063_

seeM:
This fixes a bug where notebooks were failing to preview with error 'No Quarto document available to render'.

The problem is that notebook cell text editors disappear from `vscode.window.visibleTextEditors` when either the bottom Panel or the Secondary Sidebar are visible. I'm unsure what the cause is, but it seems like a Code OSS bug or perhaps intentional change in behavior.

The fix is to stop requiring a text editor in the notebook case.

`QuartoEditor.textEditor` is only used in one place today: `PreviewManager.detectErrorNavigation` which is already skipped for notebooks, so it's safe to remove for notebooks. This allows us to remove the `visibleTextEditors` check.

Since `QuartoEditor` still requires a `TextDocument` reference, we now use the first notebook cell, if one exists.

Fixes #1006.

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## 1.134.0 (Unreleased)
 
 - The "Preview" and "Preview Format..." commands now show in the Positron Notebook Editor overflow menu (<https://github.com/quarto-dev/quarto/pull/1001>).
+- Fixed notebooks failing to render or preview when they were visible but not active, or while the Panel or Sidebar were open (<https://github.com/quarto-dev/quarto/pull/1007>).
 - In Positron, Jupyter Notebooks (`.ipynb`) are now exported via the new unified "Export" command, rather than the "Quarto: Convert to .qmd" command (<https://github.com/quarto-dev/quarto/pull/999>).
 - Fixed a bug where formatting a code cell stripped leading empty lines. Leading empty lines between option directives and code are now preserved, and two or more leading empty lines are collapsed to one (<https://github.com/quarto-dev/quarto/pull/953>).
 - Fixed a bug where IPython magics (`%`, `%%`) and shell escapes (`!`) in Python code cells produced spurious diagnostics from language servers like Pyrefly and Ruff. These lines are now commented out in the virtual document handed to language servers (<https://github.com/quarto-dev/quarto/pull/1013>).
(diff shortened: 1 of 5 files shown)


apps/vscode/src/core/doc.ts
```diff
@@ -217,47 +217,53 @@ export function findQuartoEditor(
...
+    return visibleVisualEditor;
   }
+
+  // visible notebook editors
+  const visibleNotebookEditor = vscode.window.visibleNotebookEditors.find((editor) =>
+    filter(editor.notebook.cellAt(0)?.document)
+  );
+  if (visibleNotebookEditor) {
+    const firstCellDocument = visibleNotebookEditor.notebook.cellAt(0).document;
+    return quartoNotebookEditor(visibleNotebookEditor, firstCellDocument);
```
vezwork, line 257: Consider adding a note on why we pass the first cell document.

vezwork (approved):
Nice!

apps/vscode/CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## 1.134.0 (Unreleased)
 
 - The "Preview" and "Preview Format..." commands now show in the Positron Notebook Editor overflow menu (<https://github.com/quarto-dev/quarto/pull/1001>).
+- Fixed notebooks failing to render or preview when they were visible but not active, or while the Panel or Sidebar were open (<https://github.com/quarto-dev/quarto/pull/1007>).
```
juliasilge: It looks like this file is a fair bit behind `main`. Can you merge from `main` and make sure it resolves well?

juliasilge (approved):
This is working well for me!

apps/vscode/src/core/doc.ts
seeM: Good point, added in 9630827.

apps/vscode/CHANGELOG.md
seeM: Done in 56f786b

PR #1008, opened 2026-06-15
Use Positron API npm package

seeM:
This PR updates the `hooks.ts` file to use the Positron API npm package (and renames the module and a few of its functions). I ended up doing this to fix TypeScript errors when running `yarn tsc --noEmit apps/vscode`.


apps/vscode/src/lsp/client.ts
```diff
@@ -538,21 +538,3 @@ function unadjustSymbolRanges(
...
-    // If this is not a virtual document, pass through all diagnostics
-    if (!isVirtualDoc(uri)) {
-      next(uri, diagnostics);
-      return;
-    }
-
-    // For virtual documents, filter out all diagnostics
-    next(uri, []);
-  };
-}
```
vezwork, line 558: Why was this removed?

juliasilge: The only caller of this got removed in https://github.com/quarto-dev/quarto/pull/980, I believe.

apps/vscode/src/host/index.ts
juliasilge, line 1: Can you also update the doc around line 62 in this file? Oh, and can you also update `apps/overview.md`?

juliasilge (approved):
I tried to do this one time about 6 months ago and couldn't get it to all compile and be happy, so this is so great to see! I actually thought that something was maybe wrong with the npm package itself so this is a really good outcome for us here.

apps/vscode/src/lsp/client.ts
seeM: Yeah, it was actually removed in #980. Now, it references `Diagnostic` which isn't imported anymore so it causes a TS error. I think it may have been accidentally added back after #980 in a merge conflict resolution in another PR.

apps/vscode/src/host/index.ts
seeM: Oops, done in f875507

PR #1009, opened 2026-06-15
Remove vscode types given we've upgraded vscode

seeM:
These were causing TypeScript errors when doing `yarn tsc --noEmit apps/vscode`. They may have been left behind from when we depended on an older vscode version.


vezwork:
I checked out this PR and opened up some of the files and didn't see any type errors. `npm run dev-vscode` also worked fine.

PR #1010, opened 2026-06-15
Fix typescript errors due to mismatched range type

seeM:
There was a mismatch between `Token.range`'s type (from `vscode-languageserver-types`) and the one that eventually gets passed to the Positron API. These were causing TypeScript errors when doing `yarn tsc --noEmit apps/vscode`.

### DIFF
--- a/apps/vscode/src/providers/cell/commands.ts
+++ b/apps/vscode/src/providers/cell/commands.ts
@@ -15,8 +15,9 @@
  *
  */
 
-import { lines, sleep } from "core";
-import { CodeViewActiveBlockContext, CodeViewSelectionAction } from "editor-types";
+import { lines } from "core";
+import { Range as LSRange } from 'vscode-languageserver-types';
+import { CodeViewActiveBlockContext } from "editor-types";
 import {
   Position,
   Range,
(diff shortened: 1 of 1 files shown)


apps/vscode/src/providers/cell/commands.ts
```diff
@@ -177,7 +177,7 @@ class RunCurrentCellCommand extends RunCommand implements Command {
       const executor = await this.cellExecutorForLanguage(language, editor.document, this.engine_);
       if (executor) {
         const code = codeWithoutOptionsFromBlock(block);
-        const ranges = block.range ? [block.range] : undefined;
+        const ranges = [toVscodeRange(block.range)];
```
vezwork, line 180: Consider doing `block.range as Range` instead, since this code was working before(?), we shouldn't have to change the data, just the type.

vezwork (approved):
Approved with something to consider

apps/vscode/src/providers/cell/commands.ts
seeM: Hmm I think we got lucky that it was working. The eventual caller might one day call a `vscode.Range` methods and error because it's not a `vscode.Range` instance.

vezwork: Yea, good point!

PR #1013, opened 2026-06-16
Comment out IPython magics in Python virtual documents to avoid spurious diagnostics

### LINKED ISSUE #1011: IPython magics produce third-party LSP diagnostics
Hi! I am using the Quarto extension in both VS Code and Positron.

A [Positron user](https://github.com/posit-dev/positron/discussions/14041) noticed that IPython magic lines produce incorrect diagnostics from LSP providers like Pyrefly or Ruff. For example, in a Python code cell:

```
import pandas as pd

%reload_ext pandas
```

Image

This does not happen in Jupyter notebooks, because Pyrefly uses the [ruff_notebook crate](https://github.com/astral-sh/ruff/tree/main/crates/ruff_notebook) to sanitize notebook input first, including commenting out IPython magics.

I think Quarto's vdoc should also comment out lines like this. An appropriate regex might be `/^\s*(%{1,2}|!)/` to catch line magics (%), cell magics (%%), and shell statements (!).

juliasilge:
Fixes #1011

IPython magic lines (`%`, `%%`) and shell escapes (`!`) inside Python code cells are not valid Python, so third-party language servers like Pyrefly and Ruff flag them with spurious diagnostics:

Screenshot 2026-06-16 at 10 59 08 AM

This does not happen in Jupyter notebooks, because tooling such as the `ruff_notebook` crate sanitizes notebook input first by commenting out these lines.

## Change in this PR

When Quarto builds the Python virtual document that gets fed to embedded language servers, magic and shell-escape lines are now replaced in place with the comment character (`#`) instead of being copied verbatim. This mirrors the existing mechanism that already blanks out non-code lines:

Screenshot 2026-06-16 at 10 59 26 AM

Because the lines are substituted (not removed), line counts stay identical and all the existing position-mapping logic (`adjustedLine` / `unadjustedLine`) keeps working unchanged.

Details:

- Added a `commentMagics` language option, enabled for Python only since these are IPython/Jupyter specific.
- The pattern `/^\s*(%{1,2}|!)/` (from the issue) matches line magics, cell magics, and shell escapes, including leading whitespace.
- Unlike the existing `inject` directives (`# type: ignore`, `# flake8: noqa`), this applies to the `diagnostics` action too, which is the path the false diagnostics come through.
- Added an example (`vdoc/magics.qmd`) exercising `%`, `%%`, `!`, an indented `%time`, and surrounding real Python, plus a snapshot test confirming the magic lines become `#` while real Python is preserved.
- The cell magics (`%%bash`, etc.) only comment out the magic line itself, not the rest of the cell body. This matches the issue's suggested regex and keeps the change simple; full cell-magic handling can be a follow-up if users report it.

### DIFF
--- a/apps/vscode/src/test/examples/generated_snapshots/vdoc/magics-python.py
+++ b/apps/vscode/src/test/examples/generated_snapshots/vdoc/magics-python.py
@@ -1,0 +1,25 @@
+# type: ignore
+# flake8: noqa
+#
+#
+#
+#
+#
+#
+#
+#
+import pandas as pd
+
+#
+#
+#
+#
+x = 1
+#
+#
+#
+#
+#
+#
+#
+#
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -5,6 +5,7 @@
 - The "Preview" and "Preview Format..." commands now show in the Positron Notebook Editor overflow menu (<https://github.com/quarto-dev/quarto/pull/1001>).
 - In Positron, Jupyter Notebooks (`.ipynb`) are now exported via the new unified "Export" command, rather than the "Quarto: Convert to .qmd" command (<https://github.com/quarto-dev/quarto/pull/999>).
 - Fixed a bug where formatting a code cell stripped leading empty lines. Leading empty lines between option directives and code are now preserved, and two or more leading empty lines are collapsed to one (<https://github.com/quarto-dev/quarto/pull/953>).
+- Fixed a bug where IPython magics (`%`, `%%`) and shell escapes (`!`) in Python code cells produced spurious diagnostics from language servers like Pyrefly and Ruff. These lines are now commented out in the virtual document handed to language servers (<https://github.com/quarto-dev/quarto/pull/1013>).
 
 ## 1.133.0 (Release on 2026-06-03)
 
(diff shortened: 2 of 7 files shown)


apps/vscode/src/test/examples/generated_snapshots/vdoc/magics-python.py
```diff
@@ -0,0 +1,25 @@
...
+#
+#
+#
+#
+import pandas as pd
+
+#
+#
+#
+#
```
seeM, line 16: I'm not sure if this has any user-facing consequences: should these be commented out including the original code?

```
# %reload_ext pandas
# %%timeit
# !pip install pandas
#   %time x = 1
```

seeM (approved):
Works well in my testing!

Image

apps/vscode/src/test/examples/generated_snapshots/vdoc/magics-python.py
juliasilge: I don't believe so, no. You can see how typical vdocs look here (empty comment lines):
https://github.com/quarto-dev/quarto/tree/main/apps/vscode/src/test/examples/generated_snapshots/vdoc
The extension keeps track of the mapping between the vdoc and the real doc, and the real doc is still there as well to get the info when needed.

PR #1015, opened 2026-06-18
Consolidate unique examples document helpers

seeM:
I noticed that we added two very similar helper functions to work around a caching issue. This PR consolidates those.

### DIFF
--- a/apps/vscode/src/test/test-utils.ts
+++ b/apps/vscode/src/test/test-utils.ts
@@ -80,6 +80,8 @@ export async function openUniqueExamplesDocument(fileName: string, disposables: 
   const extension = path.extname(fileName);
   const uniqueName = `${path.basename(fileName, extension)}-${Date.now()}-${Math.random().toString(36).slice(2)}${extension}`;
   const uniqueUri = vscode.Uri.joinPath(sourceUri, "..", uniqueName);
+  fs.copyFileSync(sourceUri.fsPath, uniqueUri.fsPath);
+  const doc = await vscode.workspace.openTextDocument(uniqueUri);
 
   /**
    * Ensure that the copy is deleted on dispose (usually, on test `teardown`).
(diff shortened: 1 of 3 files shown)


apps/vscode/src/test/test-utils.ts
```diff
@@ -64,22 +66,30 @@ export async function openAndShowUri(
...
+  const uniqueUri = vscode.Uri.joinPath(sourceUri, "..", uniqueName);
+
+  /**
+   * Ensure that the copy is deleted on dispose (usually, on test `teardown`).
+   * See the notes in {@link deleteDocument} for why we have to use that function.
+   */
+  disposables.add({ dispose: () => deleteDocument(doc) });
+
+  fs.copyFileSync(sourceUri.fsPath, uniqueUri.fsPath);
+  const doc = await vscode.workspace.openTextDocument(uniqueUri);
```
juliasilge: The dispose callback closes over `doc` before it's declared:

```ts
disposables.add({ dispose: () => deleteDocument(doc) });

fs.copyFileSync(sourceUri.fsPath, uniqueUri.fsPath);
const doc = await vscode.workspace.openTextDocument(uniqueUri);
```

This works at runtime, but to me it seems hard to read and reason about. 

Could we move the `disposables.add(...)` to after `const doc = ...` instead?

```ts
fs.copyFileSync(sourceUri.fsPath, uniqueUri.fsPath);
const doc = await vscode.workspace.openTextDocument(uniqueUri);

disposables.add({ dispose: () => deleteDocument(doc) });

return doc;
```

This is obviously just a nit, but what do you think?

juliasilge (approved):
Thank you for this cleanup!

apps/vscode/src/test/test-utils.ts
seeM: Ah yea good point! Done in 2cdbeb2

PR #1016, opened 2026-06-18
Fix extension tests when the repo is cloned to a long path

seeM:
I've been using worktrees named like `quarto.my-sanitized-branch-name`. This was causing tests to crash to since the user data dir was used to name a socket, and it was too long. This PR shortens the user data dir using a temp dir.


juliasilge (approved):
Working well for me!

Can you amend the commit message, so we don't get confused in the future? Or did you intend to actually switch to `.mts`? Maybe to use `satisfies` natively?

PR #1022, opened 2026-06-22
Be more discriminating in failure popups

jmcphers:
This change prevents superfluous error notifications/toasts when running cells in Positron. 

The issue was that we were not catching exceptions thrown by `executeCode`. It is appropriate for exceptions to propagate here when we can't execute the code for some structural reason, but most often the failure is caused by the code itself.

When a cell fails to execute (a runtime/syntax error, an interrupt, or the session exiting) Positron already reports it in the console. Previously the rejection from `executeCode` also propagated up to the command handler, which surfaced it a second time as a notification popup.

This change distinguishes the two failure modes via the execution observer's `onFailed` callback:

- **Runtime-reported failures** (the code ran and errored): logged at debug level and swallowed so they aren't shown twice. Subsequent blocks are skipped since one failed.
- **Submission failures** (e.g. no runtime registered for the language): logged as an error and re-thrown so the user is notified, since these aren't otherwise surfaced.

Fixes posit-dev/positron#9845

### DIFF
--- a/apps/vscode/src/host/positron.ts
+++ b/apps/vscode/src/host/positron.ts
@@ -37,7 +37,7 @@ export function isInlineOutputEnabled(): boolean {
     .get<boolean>("enabled", false);
 }
 
-export function positronExtensionHost(): ExtensionHost {
+export function positronExtensionHost(outputChannel?: vscode.LogOutputChannel): ExtensionHost {
   return {
     // supported executable languages (we delegate to the default for langugaes
     // w/o runtimes so we support all languages)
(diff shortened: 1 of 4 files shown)


apps/vscode/src/host/positron.ts
```diff
@@ -68,18 +68,52 @@ export function positronExtensionHost(): ExtensionHost {
...
+                  // failure of the running code (a runtime error or syntax
+                  // error in the cell), an interrupt, or the session exiting are
+                  // all delivered through the execution observer's `onFailed`
+                  // callback before `executeCode` rejects. These are already
+                  // surfaced in the console, so we don't want to surface them
+                  // again. A rejection *without* `onFailed` having fired means
+                  // the code couldn't be submitted to the runtime at all (e.g.
+                  // no runtime is registered for the language), which is a
+                  // genuine problem worth reporting.
+                  let runtimeFailure = false;
```
juliasilge, line 82: This logic depends on `onFailed` having already fired (synchronously setting `runtimeFailure`) by the time we reach this `catch`. The `ExecutionObserver` docs only guarantee that one of `onCompleted`/`onFailed` is called, not their ordering relative to the `executeCode` promise rejecting. If that ordering ever changed or raced, a runtime error would be misclassified as a submission failure and we'd surface the popup again. Just noting the dependency for future readers; the failure mode is only a graceful regression to the old behavior, so nothing to change here IMO.

juliasilge (approved):
This is working great! 🎉 

Image

Before you merge, can you please update the CHANGELOG with a note on this fix? You might need to merge from `main` so you get the CHANGELOG entry under the right version (we released yesterday).

apps/vscode/src/host/positron.ts
jmcphers: Yes, it's a little kludgy! I think a more complete fix might be to throw specific structured error classes from Positron to distinguish "can't run code at all" errors from "code ran but it failed" errors.

PR #1025, opened 2026-06-25
Hide the editor Preview button when Positron provides a native preview split button

juliasilge:
Positron 2026.07.0 is adding a native "Preview" split button to the editor action bar for Quarto documents as outlined in https://github.com/posit-dev/positron/issues/12891: a "Preview" primary action plus a dropdown to preview specific formats. This extension also contributes its own Preview button to `editor/title/run`, so in Positron both buttons would appear. The change in this PR hides the extension's button when Positron provides the native one, leaving a single Preview button.

It is a no-op in VS Code and in older Positron versions.

## How

- Adds a `quartoCanUsePositronPreviewSplitButton` context key, set once at activation in `context-keys.ts` when `positron.version >= "2026.07.0"`.
- Gates the `editor/title/run` `quarto.preview` contribution on `&& !quartoCanUsePositronPreviewSplitButton`.

The version check uses lexicographic string comparison rather than `semver`, matching the existing approach in `host/positron.ts`; Positron's zero-padded calendar versions such as `2026.07.0` are not valid semver, but the zero-padded year/month sort correctly as strings.

## Why a version gate rather than "is Positron"

The extension auto-updates from the marketplace independently of the Positron build, so a plain "hide whenever running in Positron" check would remove the button for users on older Positron versions that have no replacement button. Gating on the version means the button is hidden only once a Positron that actually provides the native split button is running. If that Positron release happens to slip past 2026.07.0 (which I assure you it WILL NOT 😉), the gate is a one-line follow-up.

## No behavior change elsewhere

- VS Code: no Positron API, the key is never set, the button shows as before.
- Positron < 2026.07.0: the key is not set, the button shows as before.

## Testing

- VS Code: Preview button still appears for `.qmd` / `.md` / the visual editor.
- Positron >= 2026.07.0: the extension's Preview button is hidden (Positron's native split button appears instead).
- Positron < 2026.07.0: the extension's Preview button still appears.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -728,7 +728,7 @@
       "editor/title/run": [
         {
           "command": "quarto.preview",
-          "when": "editorLangId == quarto || editorLangId == markdown || activeCustomEditorId == 'quarto.visualEditor'"
+          "when": "(editorLangId == quarto || editorLangId == markdown || activeCustomEditorId == 'quarto.visualEditor') && !quartoCanUsePositronPreviewSplitButton"
         },
         {
           "command": "quarto.previewScript",
--- a/apps/vscode/src/providers/context-keys.ts
+++ b/apps/vscode/src/providers/context-keys.ts
@@ -15,6 +15,7 @@
 
 import * as vscode from "vscode";
 import debounce from "lodash.debounce";
+import { tryAcquirePositronApi } from "@posit-dev/positron";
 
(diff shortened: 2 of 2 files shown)


juliasilge:
This is not actually the right thing to do, because it can impact `.md` files in a bad way. There is already deduping logic in the editor action bar in Positron, so we can rely on that.

PR #1027, opened 2026-06-29
Fix notebook markdown soft line breaks

### LINKED ISSUE #1017: Single soft line breaks in .ipynb markdown cells render as hard breaks in VS Code when using quarto extension
### Bug description

When the Quarto extension is enabled in VS Code, single newlines (soft breaks) within a
paragraph in a Jupyter notebook (`.ipynb`) markdown cell are rendered as hard
line breaks in the cell's rendered output. In standard CommonMark, a single
newline inside a paragraph should collapse to a space; only a blank line should
start a new paragraph.

This means hard-wrapped prose in a markdown cell renders with the wraps
preserved as `<br>`-style breaks instead of flowing as one paragraph.

The behavior is specific to Quarto: disabling the Quarto extension restores the
expected CommonMark behavior in notebook cells. The VS Code Markdown *preview*
(`.md` files) renders correctly the whole time, and I have
`"markdown.preview.breaks": false` set in my settings — that setting does not
appear to be honored by Quarto's notebook cell renderer.

### Steps to reproduce

1. Enable the Quarto extension.
2. Open a `.ipynb` file in VS Code.
3. In a markdown cell, enter text that is hard-wrapped across lines, e.g.:
 
```
Hello
world
```
 
5. Render the cell.

### Actual behavior

The soft break is preserved as a hard line break, so the output wraps after
"FSL-route" exactly where the source was wrapped.

i.e. it renders as

> Hello
> world

### Expected behavior

The cell renders as a single flowing line (the soft break collapses to a space):

I.e. 

> Hello world

### Your environment

- Quarto extension version: 1.9.38
- VS Code version: 1.125.0
- OS: macOS 26.5.1
- Relevant setting: `"markdown.preview.breaks": false`

devmcp:
Markdown cells in notebooks were rendering soft line breaks as hard paragraph breaks when the Quarto Markdown-It renderer was active. The Pandoc div rule reported every line as a paragraph interrupter during Markdown-It silent validation, so hard-wrapped prose was split before normal Markdown rendering could preserve soft break semantics.

This updates the div parser so silent validation only reports an interruption for actual `:::` div markers and does not mutate parser state until the marker is consumed. The new core tests cover both the soft-break regression and the expected paragraph interruption for Pandoc div markers.

Fixes quarto-dev/quarto#1017

Tests:
- `npx --yes yarn@1.22.22 workspace core test`
- `npx --yes yarn@1.22.22 workspace quarto-vscode-markdownit build`
- `npx --yes yarn@1.22.22 build-vscode`

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - In Positron, running a cell that raises an error no longer results in an error toast message (<https://github.com/posit-dev/positron/issues/9845>).
 - In Positron, Julia code cells in `.qmd` files can now be executed when the `ntluong95.positron-julia` extension is installed, the same way Python and R do (<https://github.com/quarto-dev/quarto/pull/989>).
+- Fixed a bug where soft line breaks within a paragraph were rendered as separate paragraphs in the markdown preview (<https://github.com/quarto-dev/quarto/pull/1027>).
 
 ## 1.134.0 (Release on 2026-06-22)
 
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -13,6 +13,10 @@
   },
   "devDependencies": {
     "@types/markdown-it": "^12.2.3",
-    "@types/markdown-it-attrs": "^4.1.0"
+    "@types/markdown-it-attrs": "^4.1.0",
+    "tsx": "^4.7.1"
+  },
+  "scripts": {
+    "test": "node --import tsx --test test/*.test.ts"
   }
 }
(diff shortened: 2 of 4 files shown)


juliasilge:
Thank you so much for this fix, @devmcp! I notice it is still draft, but it looks ready to go to me. Do you have any other changes you want to make before I review?

devmcp:
> Thank you so much for this fix, @devmcp! I notice it is still draft, but it looks ready to go to me. Do you have any other changes you want to make before I review?

Nope! I think it's good to review. Many thanks!

juliasilge (approved):
Thank you for this contribution! 🚀

PR #1033, opened 2026-07-01
Patch brace-expansion ReDoS (SNYK-JS-BRACEEXPANSION-17706650)

juliasilge:
Re-resolve the transitive brace-expansion dependency to patched versions (1.1.15, 2.1.1, 5.0.7) within each major line, closing the ReDoS advisory across all dependency paths (minimatch via vscode-languageclient, eslint, exceljs, tsup, @vscode/test-cli, and pinyin's optional nodejieba).

This supersedes the Snyk PRs (#1028, #1029, #1030, #1031), each of which proposed a major-version bump of a top-level dependency to shift a single transitive path off the vulnerable package. The lockfile re-resolution fixes every path at once with no VS Code engine bump, no eslint flat-config migration, and no pinyin segmentation change.

- Closes #1028
- Closes #1029
- Closes #1030 
- Closes #1031

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -3280,24 +3280,24 @@ boundary@^1.0.1:
   integrity sha512-AaLhxHwYVh55iOTJncV3DE5o7RakEUSSj64XXEWRTiIhlp7aDI8qR0vY/k8Uw0Z234VjZi/iG/WxfrvqYPUCww==
 
 brace-expansion@^1.1.7:
-  version "1.1.11"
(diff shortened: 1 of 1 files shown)


cscheid (approved):
Yup, all looks good from here, though I haven't recompiled or retested the extension. None of these are "real" in any concrete sense, since ReDOS is about resource exhaustion under adversarial inputs, and if people want to crash their own machines with the extension, there's much simpler ways to go about it...

PR #1043, opened 2026-07-05
Patch linkify-it ReDoS (SNYK-JS-LINKIFYIT-17817062)

juliasilge:
Force the transitive linkify-it dependency to the patched 5.0.2 via a root resolutions entry, closing the ReDoS advisory (CVE-2026-48801) across all dependency paths (markdown-it in core, quarto-core, editor-codemirror, vscode, and vscode-markdownit, plus the markdown-it bundled in the vsce packaging tool).

This supersedes the Snyk PRs (#1038, #1039, #1040, #1041, #1042), each of which proposed a major-version bump of markdown-it (12/13 to 14) to shift a single path off the vulnerable package. Unlike brace-expansion, linkify-it has no patched 3.x or 4.x release, so the fix requires linkify-it 5; the resolutions override reaches every path at once with no markdown-it 14 migration (no `@types` bump, no .mjs deep-import rewrites) and no behavior change, since linkify-it's public API is unchanged across v3 to v5.

- Closes #1038
- Closes #1039
- Closes #1040
- Closes #1041
- Closes #1042

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,6 +19,9 @@
     "format": "prettier --write \"**/*.{ts,tsx}\"",
     "watch": "concurrently 'tsc --watch --noEmit --project apps/writer/tsconfig.json' 'tsc --watch --noEmit  --project packages/writer-server/tsconfig.json'"
   },
+  "resolutions": {
+    "linkify-it": "5.0.2"
+  },
   "devDependencies": {
     "@vscode/test-cli": "^0.0.11",
     "concurrently": "^7.5.0",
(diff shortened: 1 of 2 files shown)


cscheid:
~~I think this is _quite_ risky. My understanding is that `markdownit` is the library that handles all of the non quarto-cli markdown processing in the extension.~~

~~I'm not going to have time to give this the analysis it deserves. In my estimation of risk-vs-benefit, ReDoS does not rank very highly. I understand the urgency for wanting to merge this, but I genuinely recommend against doing it before checking our exposure to major markdownit changes.~~

I apologize! Ok, I now actually understand what's going on. This seems ok. The risk here, then, if I understand correctly, is that we're forcing a dependency of linkify-it 3 to resolve to linkify-it 5, and that combination is not "usual" or typically tested. I think that's an acceptable risk; it's `markdown-it` that I'm terrified about migrating :)

cscheid (approved):

PR #1044, opened 2026-07-07
Update to handle both new and deprecated Positron setting for inline output

juliasilge:
Related to: 

- https://github.com/posit-dev/positron/pull/14742
- https://github.com/posit-dev/positron/issues/14301

We'll need to get this merged, then the extension released, and then we can finish up the Positron side work.


juliasilge:
Using the `.vsix` from the GH action here, all the E2E tests in https://github.com/posit-dev/positron/pull/14742 pass.

jmcphers (approved):
LGTM!

PR #1057, opened 2026-07-14
Patch tar DoS advisories by dropping pinyin's nodejieba (SNYK-JS-TAR-17909068, SNYK-JS-TAR-17909152)

juliasilge:
Bump `pinyin` from `^2.10.2` to `^4.0.0` in `packages/editor` and drop the now-redundant `@types/pinyin` devDependency, closing the two remaining tar advisories from #1052:

- SNYK-JS-TAR-17909068 infinite loop / CVE-2026-59874
- SNYK-JS-TAR-17909152 resource exhaustion / CVE-2026-59873

The third vulnerability in the Snyk PR, the js-yaml merge-key DoS (SNYK-JS-JSYAML-17900054), was already resolved by the lockfile re-resolution in #1056, so editor's `js-yaml: ^4.1.0` already resolves to the patched 4.3.0 and no js-yaml change is needed here.

Unlike the earlier dependency advisories, the tar fix is only available in tar 7.5.18/7.5.19; there is no patched 6.x release, so a within-major lockfile re-resolution is not possible. The vulnerable tar 6.1.13 is pulled in transitively along a single path: `editor -> pinyin -> nodejieba (optionalDependency) -> @mapbox/node-pre-gyp -> tar`. Rather than force tar 7 through node-pre-gyp 1.x (which pins `tar ^6.1.11` and is not compatible with tar 7's API/engine changes), bumping pinyin to 4.0.0 removes the entire native subtree: pinyin 4 drops the `nodejieba` optional dependency, which eliminates `nodejieba`, `@mapbox/node-pre-gyp`, and `tar` from the lockfile altogether (about 160 fewer lockfile lines and no native install-time build step).

This supersedes Snyk PR #1052, which bumped pinyin in `package.json` but did not update `yarn.lock` and left the stale `@types/pinyin` behind.

There is no behavior change from this change. The editor uses pinyin at a single call site (`createUniqueCiteId` in `cite.ts`) to romanize an individual Hanzi character into an ASCII citekey. pinyin 4 returns byte-identical romanizations for that call, its CJS build sets `__esModule` and its ESM build exports the callable as `default`, so the existing `import pinyin from 'pinyin'` default import resolves correctly under both the TypeScript esModuleInterop path and the Vite/Rollup bundle. All working correctly, shown here with a search for the DOI `10.1142/9789814285780`:

Screenshot 2026-07-14 at 10 17 52 AM

The dropped `nodejieba` only ever affected multi-character `segment` mode, which this single-character call never used. Verified with `tsc --noEmit` on `packages/editor` and a full `vscode-editor` build.

Closes #1052


cscheid (commented):
Looks good!

PR #1058, opened 2026-07-16
positron: add api tests in ci

isabelizimm:
### Summary

Adds a separate integration test suite that runs the Quarto extension inside Positron (rather than vanilla VS Code), so we can cover Positron-specific code paths that the main @vscode/test-cli suite can't reach.

### Tests added (`apps/vscode/src/test/positron/`)

- `execute-cell.test.ts` — Verifies the Quarto-specific transformations that happen before delegating to `positron.runtime.executeCode`:
  - Quarto cell options (`#| ...`) are stripped from executed code.
  - In a knitr document, Python cells are rerouted through `reticulate::repl_python(...)` and submitted to the R runtime.
- `notebook-export.test.ts` — Verifies Quarto registers its `.qmd` exporter with Positron's built-in `positron.notebook-export` extension API (cross-extension wiring absent in vanilla VS Code).

### Test harness / infra

- `scripts/run-positron-tests.mjs` — Launcher using `@posit-dev/positron-test-electron`. Runs with extensions enabled (opts out of the default `--disable-extensions`) so Positron's bundled runtimes and notebook-export extension are available. Extension auto-update is disabled in the test host — otherwise Positron treats the dev-loaded extension as "outdated" and removes it mid-run.
- `index.ts` — Mocha entry point (tdd UI, 120s timeout) that discovers and runs the compiled tests.
- `build.ts` — Includes `positron/*.ts` in the test build.
- `package.json` — Adds `test-positron` scripts (app + root) and the `@posit-dev/positron-test-electron` dev dependency.
- `.gitignore` — Ignores `.positron-test`.

Run locally with `yarn test-positron`.

### CI

- `.github/workflows/test-positron.yaml` — New workflow. Runs the tests via the `posit-dev/setup-positron` action, which downloads a Positron build and executes `yarn test-positron` inside it.


isabelizimm:
Looks like the same failure is showing up on main, but the rest should be ready to go! 

~EDIT: I'm seeing som non-deterministic failures on the vscode tests that are passing on main. I'll dig into them more tomorrow!~ was a flake 😩 passing now!

juliasilge:
Failing tests are related to https://github.com/quarto-dev/quarto/pull/1062, I think

juliasilge:
The fix for the tests related to the new Quarto version is merged now, if you want to merge from `main` and see if there is anything else to do here!

isabelizimm:
Ah, yep! Looks much better now 💯

juliasilge (approved):
Thank you so much for working on these! We're in a much better place now for being able to add tests moving forward.

vezwork:
Sorry I missed this PR before it was merged. It looks awesome @isabelizimm !! Excited to have this.

PR #1060, opened 2026-07-20
don't intercept pastes inside citation ids

kevinushey:
Addresses https://github.com/rstudio/rstudio/issues/18295

## Problem

In the visual editor on Windows, typing `@` and pasting a DOI (e.g. `https://doi.org/10.1000/182`) does nothing, whereas on Linux/macOS the editor offers to look up the DOI and insert a citation.

## Cause

ProseMirror consults `handlePaste` plugins in registration order, and the generic paste extension (`behaviors/paste.ts`) is registered before the cite mark's `paste_cite_doi` plugin. Inside a `cite_id` mark the generic handlers form a pincer that prevents `paste_cite_doi` from ever running on Windows:

- Clipboards carrying `text/html` (e.g. a DOI copied from a browser page) are consumed by `pasteHtmlHandler`'s `isWindows()` branch (the workaround for ProseMirror freezing on multi-paragraph pastes).
- Plain-text-only clipboards are consumed by `pasteMarkdownHandler` (the cite marks don't set `noInputRules`, so `allowMarkdownPaste` returns true inside a citation). This variant also affects macOS and Linux.

Since `isWindows()` keys off the user agent, this affects any Windows client, including browsers connecting to a Linux server.

## Fix

Both generic handlers now return `false` when the selection is inside a `cite_id` mark, allowing the cite mark's own paste handler to process the paste (insert the DOI, offer to insert a citation, or fall back to a plain-text insert for non-DOI content).

## Tests

Added `apps/vscode/src/test/editor-paste.test.ts` covering the fixed cases plus the preserved behaviors (Office content and Windows `text/html` pastes outside citations are still handled, plain text outside citations is still pasted as markdown). The 3 regression cases fail without the fix.

The tests live in the vscode extension suite (rather than `packages/editor`, whose jest-style tests under `test/api` are orphaned -- no runner is configured for them, and `test/README.md` notes their fixtures no longer work) because it is the only harness exercised by CI, and this avoids adding a new test runner dependency. The tests exercise the editor package directly and don't use the vscode API; the only dependency change is adding the `editor` workspace package as a devDependency of the extension.

Verified `tsc --noEmit` and the panmirror bundle build pass. Manual verification on Windows is still pending.


apps/vscode/src/test/editor-paste.test.ts
```diff
@@ -0,0 +1,157 @@
...
+const kDOIText = "https://doi.org/10.1000/182";
+const kDOIHtml = '<a href="https://doi.org/10.1000/182">https://doi.org/10.1000/182</a>';
+const kWordHtml = '<html xmlns:w="urn:schemas-microsoft-com:office:word"><body><p>content</p></body></html>';
+
+suite("Editor Paste Handlers", function () {
+  suiteTeardown(function () {
+    if (kNavigatorDescriptor) {
+      Object.defineProperty(globalThis, "navigator", kNavigatorDescriptor);
+    } else {
+      Reflect.deleteProperty(globalThis, "navigator");
```
vezwork, line 114: I think this will leave the globalThis without a navigator property? Should we rather store `globalThis.navigator` in a variable e.g. `originalNavigator`, and then later do `Object.defineProperty(globalThis, "navigator", originalNavigator);`?

vezwork: Not sure what happened here, but I meant to comment on lines 105 thru 108.

vezwork (approved):
The tests look reasonable to me, I left one comment. Ran the tests locally on mac and they passed. I checked out this PR and on Mac VSCode I tried out typing @ and pasting a DOI in the visual editor, and it successfully searched for a DOI; So the functionality seems unaffected on Mac, as it should be.

I'm giving my approval, but I do think you should await manual verification on windows before merging @kevinushey .


## An aside

I believe this to either be my incompetence or a separate issue, but: this functionality seems to always fail to look up the DOI, screenshot attached. Thought I'd put that here in case its somehow relevant.

Image

apps/vscode/src/test/editor-paste.test.ts
kevinushey: I believe this case is handled: `kNavigatorDescriptor` is captured at module load, before any `setUserAgent()` call, so it plays the role of `originalNavigator` here. If `navigator` existed originally, the `if` branch restores its exact descriptor; the `delete` branch only runs when there was no `navigator` to begin with (older Node versions), so removing the property is the correct restoration there.

One reason to prefer saving the descriptor over the value: in newer versions of Node, `globalThis.navigator` is defined as a lazy getter, and restoring the saved descriptor preserves that, whereas re-defining with the saved value would convert it to a plain data property.

vezwork: Ah, I misunderstood the code. I see now, thank you! Looks good to me

kevinushey:
I think the issue you were seeing was because the tests used the canonical example DOI (10.1000/182, the DOI Handbook), which resolves to a landing page but doesn't have citation metadata behind it -- so the lookup gets HTML instead of CSL JSON. I've updated the tests to use a real Crossref DOI and left a comment explaining this.

vezwork:
When I type @ and paste the updated DOI (10.1038/nature12373) it gets looked up properly now! Thanks

vezwork:
Hmmmm something seems to be going wrong with the basic roundtrip test in the Quarto pre-release @cscheid . Could this have to do with the new pandoc changes?

vezwork:
@cderv has a PR up to fix the failing pre release test: https://github.com/quarto-dev/quarto/pull/1062 (to be merged after Quarto is released). 

Please ignore the failing pre release test.

PR #1062, opened 2026-07-24
Update roundtrip snapshots pandoc 3.10

cderv:
## What

Updates the two visual-editor roundtrip snapshots affected by Quarto's move to Pandoc 3.10 (pre-release):

- `roundtripped-valid-basics-2.qmd`
- `roundtripped-valid-nesting.qmd`

## Why

Pandoc 3.10 changes two Markdown-writer behaviors these snapshots exercise:

1. **Display math** — Pandoc 3.7 started collapsing block math `$$…$$` onto a single line (jgm/pandoc#11384). It was reverted in Pandoc 3.9 (jgm/pandoc@8123be6), restoring the pre-3.7 multi-line form. Quarto jumped from Pandoc 3.6.3 to 3.8.3 in 1.9.x — skipping 3.9 — so the snapshots were   updated in #944 to the collapsed 3.8.3 output. Pandoc 3.10 reverts that, so the math here goes back to multi-line. Note this is **identical to the original snapshots added in #790** (which were generated with Pandoc 3.6.3).

2. **Definition lists** — with `four_space_rule` disabled (Quarto's default), Pandoc 3.10 indents definition content by one space instead of three (jgm/pandoc@8ce63ff). This is a new, intended change (not a revert). Pandoc deactivated four space rule in https://github.com/jgm/pandoc/commit/add83e81698a3468b4fde02f3af3c42d5c56f790 but missed a case for definition list I believe. 

The table change in `valid-nesting` is a consequence of (1): the display-math cell reflows, which narrows the column.

## Notes

- Alternative for the definition-list change: enable `+four_space_rule` in the visual editor's Markdown writer format to keep 4-space indentation stable across Pandoc versions. Left out here since it changes real output, not just tests. And I don't think we want this. 

- gitignore addition is because I use roborev and this is needed


cderv:
Probably to merge once pre-release becomes release, as test is failing now on Release only (which is expected)

juliasilge:
This is passing now! I am going to merge so we can evaluate tests in other PRs more easily.

PR #1063, opened 2026-07-25
Fix cell/outline breakage after single-line display math with attributes

### LINKED ISSUE #976: Quarto extension display math edge cases stops extension functionality for rest of document
### Bug description

There appears to be a problem with parsing documents after a display math environment that:

1. Includes the `$$` characters on the same line as the equation, *and*
2. has a reference label.

After one of these, the Quarto extension:

1. No longer includes headers in the Outline view in Positron;
2. Does not show the Preview button for that or subsequent display math environments; and
3. Does not recognize any code chunks after the display math environment.

This seems like it might be related to #521 .

### Steps to reproduce

Open the following Gist in Positron:

[https://gist.github.com/johnbest-wdfw/596c2843b03bfc36cef0eac103b04c16](https://gist.github.com/johnbest-wdfw/596c2843b03bfc36cef0eac103b04c16)

Observe the behavior in the editor.

### Actual behavior

After one or two equations with `$$` fences on the same line as the equation and a reference label, section headers and code chunks no longer populate the Outline pane, LaTeX Preview is no longer available, and code chunks cannot be sent to the console, giving an "Editor selection is not within an executable cell" error.

### Expected behavior

Section headers and code chunks should always populate the Outline pane, LaTeX Preview should always be available, and code chunks should always be able to be sent to the console.

### Your environment

Positron Version: 2026.05.1 (user setup) build 2
Quarto extension version: 1.132.0

Code - OSS Version: 1.109.2
Commit: 80db96acd1bc6f7e7ff5ba4cbc509af8d40af1dd
Date: 2026-05-05T18:02:55.030Z
Electron: 39.3.0
Chromium: 142.0.7444.265
Node.js: 22.21.1
V8: 14.2.231.22-electron.0
OS: Windows_NT x64 10.0.26100

max-nothacker:
Fixes #976

## Problem

A single-line display math expression with a cross-reference label, e.g.

```markdown
$$1+1$$ {#eq-spec0}
```

breaks the extension for the rest of the document: headings disappear from the Outline, LaTeX preview buttons vanish, and later code cells lose their Run Cell button ("Editor selection is not within an executable cell").

## Root cause

In `math_block()` (`packages/core/src/markdownit/math.ts`), the single-line close check was

```ts
if (firstLine.trim().slice(-2) === "$$") {
```

which fails when an attribute block trails the closing `$$`. The rule then falls through to the multi-line scan looking for `/^\$\$\s*(\{.*\})?\s*$/` on a later line — which never matches — so the loop runs to the end of the document and the math token swallows everything after the equation. Nothing downstream (headings, fences) is tokenized again. Full analysis in https://github.com/quarto-dev/quarto/issues/976#issuecomment-5078536895.

## Fix

Recognize an optional trailing attribute block in the single-line case, mirroring the multi-line close regex, and store it in `token.info` the same way the multi-line path does (so single-line labeled equations also pick up the math preview decoration):

```ts
let attrStr = undefined;
const singleLineMatch = firstLine.trim().match(/^(.*)\$\$\s*(\{.*\})?\s*$/);
if (singleLineMatch) {
  firstLine = singleLineMatch[1];
  attrStr = singleLineMatch[2];
  found = true;
}
```

Unlabeled single-line math (`$$1+1$$`) takes the same path it did before (`(.*)` is greedy, so interior `$$` behave as with the previous `slice(-2)` check), and the multi-line branch is untouched.

## Tests

Added `packages/core/test/markdownit-math.test.ts` (same harness as the existing `markdownit-divs.test.ts`):

1. single-line labeled equation followed by a `{python}` fence → one `math_block` with `info === "{#eq-spec0}"`, fence still tokenized (the #976 symptom)
2. single-line unlabeled equation → unchanged behavior
3. multi-line equation with label on the closing line → unchanged behavior

`yarn test` in `packages/core` passes (5/5), and `tsc --noEmit` is clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code) in the session with id 6b8b182e


juliasilge (approved):
Thank you so much for this contribution, @max-nothacker! 🙌 

I made a few additional changes to `math_block` and also set up CI so these tests actually get run, so we protect against breakages here in the future.

PR #1087, opened 2026-08-12
Add reflow comments in cell command

### LINKED ISSUE #408: Add VS Code support for a rewrap or reflow action for comments
If I have a Quarto file like this:

````md
---
title: "About"
---

It is a truth universally acknowledged, that a single man in possession of a good fortune must be in want of a wife.

```{r}
# It is a truth universally acknowledged, that a single man in possession of a good fortune must be in want of a wife.
# 
# However little known the feelings or views of such a man may be on his first entering a neighbourhood, this truth is so well fixed in the minds of the surrounding families, that he is considered as the rightful property of some one or other of their daughters.
# 
# "My dear Mr. Bennet," said his lady to him one day, "have you heard that Netherfield Park is let at last?"
# 
# Mr. Bennet replied that he had not.
# 

1 + 1 
```
````

then in RStudio, I can use "Reflow comment" to break up the comments into my chosen line length. I see that VS Code has [this extension](https://open-vsx.org/extension/stkb/rewrap) but it does not work at all in a Quarto document (does nothing).

### PR DESCRIPTION
**Pull request by vezwork**
Fixes https://github.com/quarto-dev/quarto/issues/408

https://github.com/user-attachments/assets/4a6a5289-0471-471e-a810-3d6e3156387b

- Adds a command "Quarto: Reflow Comments in Cell" that breaks up long comment lines into multiple comment lines (by default 80 characters is considerered long, but that is configurable by `quarto.cells.reflowColumn`)

Tested lightly in both VSCode and Positron.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -672,7 +672,7 @@
         },
         {
           "command": "quarto.reflowCommentInCell",
-          "when": "editorLangId == quarto && !editorHasSelection",
+          "when": "editorLangId == quarto",
           "group": "1_modification"
         },
         {
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -379,7 +379,7 @@
       },
       {
         "command": "quarto.reflowCommentInCell",
-        "title": "Reflow Comment in Cell",
+        "title": "Reflow Comments in Cell",
         "category": "Quarto"
       },
       {
@@ -1037,7 +1037,7 @@
           "scope": "window",
           "type": "number",
           "default": 80,
-          "markdownDescription": "Maximum line length used by the **Quarto: Reflow Comment in Cell** command."
+          "markdownDescription": "Maximum line length used by the **Quarto: Reflow Comments in Cell** command."
         },
         "quarto.cells.background.enabled": {
           "type": "boolean",
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -378,6 +378,11 @@
         "category": "Quarto"
       },
       {
+        "command": "quarto.reflowCommentInCell",
+        "title": "Reflow Comment in Cell",
+        "category": "Quarto"
+      },
+      {
         "command": "quarto.previewMath",
         "category": "Quarto",
         "title": "Preview Math"
@@ -662,6 +667,11 @@
       "editor/context": [
         {
           "command": "quarto.formatCell",
+          "when": "editorLangId == quarto && !editorHasSelection",
+          "group": "1_modification"
+        },
+        {
+          "command": "quarto.reflowCommentInCell",
           "when": "editorLangId == quarto && !editorHasSelection",
           "group": "1_modification"
         },
@@ -1021,6 +1031,13 @@
           "type": "number",
           "default": 500,
           "markdownDescription": "Delay in milliseconds before updating diagnostics after document changes."
+        },
+        "quarto.cells.reflowColumn": {
+          "order": 28,
+          "scope": "window",
+          "type": "number",
+          "default": 80,
+          "markdownDescription": "Maximum line length used by the **Quarto: Reflow Comment in Cell** command."
         },
         "quarto.cells.background.enabled": {
           "type": "boolean",
--- a/apps/vscode/src/providers/cell/reflow.ts
+++ b/apps/vscode/src/providers/cell/reflow.ts
@@ -28,7 +28,7 @@ import { Command } from "../../core/command";
 import { isQuartoDoc } from "../../core/doc";
 import { MarkdownEngine } from "../../markdown/engine";
 import { languageFromBlock } from "../../vdoc/vdoc";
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

File: apps/vscode/src/providers/cell/reflow.ts [hunk 1]
```diff
@@ -0,0 +1,273 @@
...
+    const document = editor.document;
+    if (!isQuartoDoc(document)) {
+      window.showInformationMessage("Active editor is not a Quarto document");
+      return;
+    }
+
+    const includeFence = false;
+
+    const tokens = this.engine_.parse(document);
+    const block = languageBlockAtLine(tokens, editor.selection.start.line, includeFence);
```
juliasilge, line 59: Can this use the selection when there is one? Today it reflows every comment run in the cell, so a user who selects three comment lines still gets the whole cell rewrapped. Filtering `reflows` to the runs that intersect the selection would cover it. The context menu needs a change too: the `when` clause at `apps/vscode/package.json:675` hides the command while a selection exists.

juliasilge (changes requested):
This is working well for most situations where I think people would want it! I ran `reflowComments` directly on some other comment shapes, and IMO it rewrites a few things people will not expect. The command walks the _whole_ cell body, so the surprising edits are not always near the cursor. One invocation on this cell produces two separate edits:

````
```{r}
# This is a fairly long comment about the join that goes well past the eighty column mark and so it gets rewrapped
x <- 1

# TODO before release:
# - check the join keys
# - drop the temp columns
```
````

The first comment wraps, which is what the user asked for. The TODO list four lines below also collapses, into `# TODO before release: - check the join keys - drop the temp columns`.

Shapes that come out changed:

- Bullet lists inside a comment join into one paragraph.
- `#!/usr/bin/env bash` becomes `# !/usr/bin/env bash`. The prefix fallback in `parseLine` moves the `!` into the text.
- A cell indented inside a list item loses its indentation, because markdown-it returns dedented text.
- A line that starts with `#` inside a multi-line string gets wrapped as though it were a comment.

None of these break a render and each one is a single undo, so I don't think the heuristic needs to learn every structure. How about we make an escape hatch? When there is a selection, reflow only the comment runs _inside_ the selection. Then a user who hits one of these cases can select the paragraphs they want and leave the rest of the cell alone. I included an inline comment where I think this should happen.

File: apps/vscode/src/providers/cell/reflow.ts (hunk 1 above)
Comment (vezwork) [reply apps/vscode/src/providers/cell/reflow.ts:59]: This seems like a good idea.

**Comment by vezwork** [comment]
@juliasilge Thanks for the review! Those are nice examples for me to consider... I wonder, what would you think about changing the command behaviour so that it only splits long comment lines, never joins them? It could also not normalize lines that aren't split e.g. to avoid cases like where `#!/usr/bin/env bash` becomes `# !/usr/bin/env bash`. 

I think that would address the examples you gave, except:

> - A line that starts with # inside a multi-line string gets wrapped as though it were a comment.

I think we'd have to look at token information to avoid this case... idk how to do this but I will look into it.

Alternatively, we could make it so the command only reflows a comment if it overlaps the cursor? Maybe that gives the user the most control of when to apply this tool to the point that these strange cases don't matter as much? But would it be tedious?

**Comment by juliasilge** [comment]
> only splits long comment lines, never joins them

That seems like a good option to me!

> we could make it so the command only reflows a comment if it overlaps the cursor

That would be an option, yes, or we could consider two commands if you don't want the command to behave differently in different situations:

- _Reflow Comments in Cell_
- _Reflow Comments in Selection_

**Comment by vezwork** [comment]
@juliasilge thanks for the input. The command now only splits long comment lines, never joins them. I missed your original proposal:

> When there is a selection, reflow only the comment runs inside the selection.

I prefer your proposal over adding two commands because it feels more intuitive and powerful to me. In the implementation, I re-interpreted your proposal a bit to be: when there is a selection, only reflow comment lines overlapping the selection.

Tested in Vscode and Positron. Updated tests. 

Please re-review!

File: apps/vscode/package.json [hunk 2]
```diff
@@ -665,6 +670,11 @@
           "when": "editorLangId == quarto && !editorHasSelection",
           "group": "1_modification"
         },
+        {
+          "command": "quarto.reflowCommentInCell",
+          "when": "editorLangId == quarto && !editorHasSelection",
```
juliasilge: This `when` clause still has `!editorHasSelection`. When a selection exists, the context menu hides the command. As a result, the new selection behavior is not reachable from the right-click menu.

I think we need to remove the clause for this command:

```suggestion
          "when": "editorLangId == quarto",
```

juliasilge (approved):
This is working great for me! 🎉 I have one change that I still think we need to make, but no need to ping me for an additional review if that change makes sense to you too.

apps/vscode/package.json
vezwork: Ah! I haven't been testing the right-click menu. I didn't even know you could run these commands from right-click!

PR #1088, opened 2026-08-13
#| highlighting alternative

### LINKED ISSUE #409: Add VS Code support to distinguish hash pipe comments from regular comments

vezwork:
Fixes #409 

This is an alternative approach to https://github.com/quarto-dev/quarto/pull/1084.

Screenshot 2026-08-13 at 5 06 14 PM

This PR is a very simple change: it adds a yaml injection language for #| comments. This means that our syntax highlighting will delegate to the yaml language highlighter inside of #| comments.

The quarto.tmLanguage change is generated from the build-lang.js script.

IMO this is a more "correct" approach, if you will. We lose control over how the yaml looks, but we shouldn't really have that control anyway (let yaml be yaml)?

### Pros
- very simple
- uses VSCode's yaml highlighting so it'll match frontmatter (which uses the same injection language approach) and yaml files

### Cons 

- yaml highlighting is a little visually intense in default themes
- no background colour
- possibly has some weird edge cases? I expected multiline yaml stuff to not work well, but it seems to work just fine. I tried using some yaml-feature-test yaml as a hash pipe comment, and it worked well.

### Testing note

We could test that the generated tmLanguage delegates `#|` properly in some examples, in a non-integrated way, by using `vscode-textmate` and `vscode-oniguruma` as dependencies in the tests, but that seems a little heavy handed so I didn't do that.

### DIFF
--- a/apps/vscode/syntaxes/build-lang.js
+++ b/apps/vscode/syntaxes/build-lang.js
@@ -491,6 +491,13 @@ const fencedCodeBlockDefinition = (
       while: (^|\\G)(?!\\s*([\`~]{3,})\\s*$)
       contentName: ${contentName}
       patterns:
+        - begin: ^(\\s*)(#\\|)
+          while: ^(\\s*)(#\\|)
+          captures:
+            '2': {name: 'comment.line.number-sign.quarto'}
+          contentName: meta.embedded.block.yaml
+          patterns:
+            - {include: 'source.yaml'}
 ${indent(4, scopes)}
 `;
 };
--- a/apps/vscode/syntaxes/quarto.tmLanguage
+++ b/apps/vscode/syntaxes/quarto.tmLanguage
@@ -173,6 +173,29 @@
             <key>patterns</key>
             <array>
               <dict>
+                <key>begin</key>
+                <string>^(\s*)(#\|)</string>
+                <key>while</key>
+                <string>^(\s*)(#\|)</string>
+                <key>captures</key>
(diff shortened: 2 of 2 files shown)


vezwork:
@mine-cetinkaya-rundel what do you think of this approach #1088  to `#|` highlighting v.s. https://github.com/quarto-dev/quarto/pull/1084#issuecomment-5269648201? The highlights (pun intended) are that this approach is simpler and uses built-in yaml highlighting so it'll be consistent across the IDE, but gives us less control so we can't i.e. have a background behind the `#|` comment lines.

vezwork:
whoops, accidentally merged this branch into main somehow. Gonna undo that.

mine-cetinkaya-rundel:
@vezwork Even though there's no background color, there's still a line separating the yaml code cell options and the code, is that correct? If so, that's pretty good still for the visiual differentiation between the two pieces of the code cell.

vezwork:
@mine-cetinkaya-rundel there's no line. Actually, let me look into combining these two approaches. Maybe we can use built-in yaml highlighting and also add a background.

vezwork:
This PR was accidentally merged to `main` and then [reverted](https://github.com/quarto-dev/quarto/commit/985e4717c2b88ca1a877be1b6cf80e41874db73b). The PR continues in https://github.com/quarto-dev/quarto/pull/1092. Sorry for confusion

PR #1092, opened 2026-08-14
#| highlighting alternative

### LINKED ISSUE #409: Add VS Code support to distinguish hash pipe comments from regular comments

vezwork:
Fixes #409 

This is an alternative approach to https://github.com/quarto-dev/quarto/pull/1084.

This is a duplicate of https://github.com/quarto-dev/quarto/pull/1088, which I accidentally merged to main and then reverted. Sorry for confusion.

Screenshot 2026-08-14 at 10 57 15 AM
Screenshot 2026-08-14 at 11 01 56 AM

This PR is a very simple change: it adds a yaml injection language for #| comments. This means that our syntax highlighting will delegate to the yaml language highlighter inside of #| comments.

The quarto.tmLanguage change is generated from the build-lang.js script.

IMO this is a more "correct" approach, if you will. We lose control over how the yaml looks, but we shouldn't really have that control anyway (let yaml be yaml)?

### Pros
- very simple
- uses VSCode's yaml highlighting so it'll match frontmatter (which uses the same injection language approach) and yaml files

### Cons 

- yaml highlighting is a little visually intense in default themes (I attempted to address this by setting text opacity to 0.75 using a decoration. I have mild worries about contrast accessibility, perhaps we should add some configuration around this.).
- possibly has some weird edge cases? I expected multiline yaml stuff to not work well, but it seems to work just fine. I tried using some yaml-feature-test yaml as a hash pipe comment, and it worked well.

### Testing note

We could test that the generated tmLanguage delegates `#|` properly in some examples, in a non-integrated way, by using `vscode-textmate` and `vscode-oniguruma` as dependencies in the tests, but that seems a little heavy handed so I didn't do that.

### DIFF
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -1067,6 +1067,13 @@
           "default": 250,
           "markdownDescription": "Millisecond delay between background color updates."
         },
+        "quarto.cells.options.background": {
+          "order": 23,
+          "scope": "window",
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Apply a visual treatment to cell option comments (e.g. `#|` lines): a slightly darker background, dimmed text, and a separator below the options."
+        },
         "quarto.cells.useReticulate": {
           "order": 27,
           "scope": "window",
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -1072,7 +1072,7 @@
           "scope": "window",
           "type": "boolean",
           "default": true,
-          "markdownDescription": "Apply a visual treatment to cell option comments (e.g. `#|` lines): a slightly darker background, dimmed text, and a separator below the options."
+          "markdownDescription": "Apply a visual treatment to cell option comments such as `#|` lines: a slightly darker background, dimmed text, and a separator below the options. When `#quarto.cells.background.color#` is `off`, this setting has no effect."
         },
         "quarto.cells.useReticulate": {
           "order": 27,
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -1495,7 +1495,7 @@
     "build": "tsx build.ts",
     "dev": "yarn run build dev",
     "lint": "eslint src --ext ts",
-    "build-lang": "node syntaxes/build-lang",
+    "build-lang": "tsx syntaxes/build-lang.js",
     "build-test": "yarn run build test",
     "test": "yarn build-test && vscode-test",
     "test-positron": "yarn build-test && node ./scripts/run-positron-tests.mjs"
--- a/apps/vscode/package.json
+++ b/apps/vscode/package.json
@@ -995,28 +995,28 @@
           ]
         },
         "quarto.cells.hoverHelp.enabled": {
-          "order": 23,
+          "order": 24,
           "scope": "window",
           "type": "boolean",
           "default": true,
           "markdownDescription": "Show help when hovering over functions."
         },
         "quarto.cells.signatureHelp.enabled": {
-          "order": 24,
+          "order": 25,
           "scope": "window",
           "type": "boolean",
           "default": true,
           "markdownDescription": "Show parameter help when editing function calls."
         },
         "quarto.cells.diagnostics.enabled": {
-          "order": 25,
+          "order": 26,
           "scope": "window",
           "type": "boolean",
           "default": true,
           "markdownDescription": "Enable diagnostics (linting) for code blocks from language servers."
         },
         "quarto.cells.diagnostics.debounceDelay": {
-          "order": 26,
+          "order": 27,
           "scope": "window",
           "type": "number",
           "default": 500,
@@ -1075,7 +1075,7 @@
           "markdownDescription": "Apply a visual treatment to cell option comments such as `#|` lines: a slightly darker background, dimmed text, and a separator below the options. When `#quarto.cells.background.color#` is `off`, this setting has no effect."
         },
         "quarto.cells.useReticulate": {
-          "order": 27,
+          "order": 28,
           "scope": "window",
           "type": "boolean",
           "default": true,
--- a/apps/vscode/src/providers/background.ts
+++ b/apps/vscode/src/providers/background.ts
@@ -212,6 +212,7 @@ async function setEditorHighlightDecorations(
 
 function clearEditorHighlightDecorations(editor: vscode.TextEditor) {
   editor.setDecorations(highlightingConfig.backgroundDecoration(), []);
+  editor.setDecorations(highlightingConfig.inlineBackgroundDecoration(), []);
   editor.setDecorations(cellOptionsBackgroundDecoration, []);
   editor.setDecorations(cellOptionsSeparatorDecoration, []);
 }
--- a/apps/vscode/src/providers/background.ts
+++ b/apps/vscode/src/providers/background.ts
@@ -153,13 +153,28 @@ async function setEditorHighlightDecorations(
   // ranges to highlight
   const blockRanges: vscode.Range[] = [];
   const inlineRanges: vscode.Range[] = [];
+  const optionLineRanges: vscode.Range[] = [];
+  const optionSeparatorRanges: vscode.Range[] = [];
 
   if (highlightingConfig.enabled()) {
 
     // find code blocks
     const tokens = engine.parse(editor.document);
     for (const block of tokens.filter(isExecutableLanguageBlock)) {
-      blockRanges.push(vscRange(block.range));
+      const blockRange = vscRange(block.range);
+      blockRanges.push(blockRange);
+
+      // cell options (#| comments) get a darker background, and the last
+      // option line gets a separator (rendered as a bottom border)
+      const lines = hashPipeLines(editor.document, blockRange);
+      for (const line of lines) {
+        optionLineRanges.push(editor.document.lineAt(line).range);
+      }
+      if (lines.length > 0) {
+        optionSeparatorRanges.push(
+          editor.document.lineAt(lines[lines.length - 1]).range
+        );
+      }
     }
 
     // find inline executable code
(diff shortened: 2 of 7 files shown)


apps/vscode/src/providers/background.ts
```diff
@@ -186,10 +201,75 @@ async function setEditorHighlightDecorations(
...
+// for c, etc.), derive the prefix from the block's language using
+// kLangCommentChars/optionCommentPattern in packages/core/src/jupyter/options.ts
+function hashPipeLines(
+  document: vscode.TextDocument,
+  blockRange: vscode.Range
+): number[] {
+  const lines: number[] = [];
+  const lastLine = Math.min(blockRange.end.line, document.lineCount - 1);
+  for (let i = blockRange.start.line + 1; i <= lastLine; i++) {
+    if (!/^\s*# ?\|/.test(document.lineAt(i).text)) {
```
juliasilge: This regex does not match the same lines as Quarto's own option parser. That parser uses `optionCommentPattern` in `cell/options.ts`, which expands to `/^#\s*\| ?/` for `#` languages and accepts any whitespace between `#` and `|`.

`format.ts` already imports `optionCommentPattern` from `cell/options.ts` so that its code path cannot drift from the option parser. Can we import it here too, instead of making a third copy of this pattern?

apps/vscode/syntaxes/build-lang.js
```diff
@@ -491,6 +491,13 @@ const fencedCodeBlockDefinition = (
       while: (^|\\G)(?!\\s*([\`~]{3,})\\s*$)
       contentName: ${contentName}
       patterns:
+        - begin: ^(\\s*)(#\\|)
```
juliasilge: This rule matches only `#|`, but the engine also accepts interior spaces (`optionCommentPattern` expands to `/^#\s*\| ?/`), and the decoration in `background.ts` accepts `# |`. A line like `# | echo: false` gets the darker background but no YAML colors. Can we make this pattern match the same lines as `optionCommentPattern`?

apps/vscode/syntaxes/build-lang.js
```diff
@@ -491,6 +491,13 @@ const fencedCodeBlockDefinition = (
...
+          while: ^(\\s*)(#\\|)
```
juliasilge: This rule fires on any `#|` run in the cell body, not only the leading run, but Quarto only reads options from the leading run, and the new decoration in `background.ts` stops at the first non-option line. Is it possible for us to handle this the same way Quarto itself does?

````
```python
x = 1
#| eval: false
```
````

Maybe not with the declarative TextMate approach, in which case let's just document it.

File: apps/vscode/src/providers/background.ts [hunk 4]
```diff
@@ -186,10 +201,75 @@ async function setEditorHighlightDecorations(
     highlightingConfig.inlineBackgroundDecoration(),
     inlineRanges
   );
+  editor.setDecorations(cellOptionsBackgroundDecoration, optionLineRanges);
+  editor.setDecorations(cellOptionsSeparatorDecoration, optionSeparatorRanges);
 }
 
 function clearEditorHighlightDecorations(editor: vscode.TextEditor) {
   editor.setDecorations(highlightingConfig.backgroundDecoration(), []);
+  editor.setDecorations(cellOptionsBackgroundDecoration, []);
```
juliasilge, line 218: This clears the block decoration and both new cell option decorations, but not `highlightingConfig.inlineBackgroundDecoration()`. When the document's language mode changes, the inline backgrounds for `` `r ...` `` code stay behind on the non-Quarto document. That omission existed _before_ this PR, but since we are adding clear calls here anyway, can we fix it? Just one line to do so, I believe.

apps/vscode/src/providers/background.ts
```diff
@@ -186,10 +201,75 @@ async function setEditorHighlightDecorations(
...
+
+// document lines of the leading run of cell option (#|) comments in a cell
+// (same pattern as the tmLanguage rule in ../../syntaxes/build-lang.js:
+// optional leading whitespace, then "#|" or "# |")
+//
+// note: this only handles #-comment languages (r, python, julia, etc.).
+// to generalize to all languages (//| for js, --| for sql, /*| ... */
+// for c, etc.), derive the prefix from the block's language using
+// kLangCommentChars/optionCommentPattern in packages/core/src/jupyter/options.ts
+function hashPipeLines(
```
juliasilge: Building on the `optionCommentPattern` comment I added, the note at lines 256-259 defers `//|` and `--|` support, but the pieces already exist and we can bring them in a pretty straightforward way. The block token carries the language (`languageNameFromBlock`, exported from quarto-core), and `kLangCommentChars` in `cell/options.ts` maps each language to its comment prefix. It _is_ private now, so it needs an export. With those two, this function works for js, sql, lua, and ojs cells too, and this third copy of "what is an option line" goes away.

apps/vscode/syntaxes/build-lang.js
juliasilge: Also, this rule is stamped into every embedded language, but `#|` is only the option prefix for `#` comment languages. Cells in other languages use `--|` (sql, lua), `//|` (js, ojs, cpp), or `%|` (matlab) and get no YAML highlighting. The reverse also happens: a literal `#|` line in a css, json, or yaml block is mis-scoped as active YAML. The generator already emits one definition per language, so we can take the prefix per language (see `kLangCommentChars` in `packages/core/src/jupyter/options.ts`).

juliasilge (changes requested):
I really like this approach, and I prefer it to #1084!

Image

I have some inline comments so we are more consistent about what the special comments are, but I'd also like to ask that we make that background/underline/opacity treatment configurable. Can we add a configuration point similar to `quarto.cells.background.color` here so that folks can turn this off/on? I think boolean and default on is fine, but I would like to give folks a way to opt out of that visual treatment.

apps/vscode/src/providers/background.ts
vezwork: hmmm. I see what you are saying but I am a bit confused: I played around with `quarto render` and the preview, and it seems that neither rendering nor previewing respect hash-pipe comments with spaces between i.e. `# |`. I was trying out:

```{r}
# | echo: false

1 + 1
```
which still echoes the cell, v.s.
```{r}
#| echo: false

1 + 1
```
which does not echo.

Am I missing something? Or is it the case that we should modify `optionCommentPattern` to be consistent with how preview and render work?

vezwork: Ok I looked into it a bit more and I now understand that `knitr` doesn't respect a space between (`# |`) but otherwise Quarto does. I guess I will not modify `optionCommentPattern`.

vezwork: We are now reusing `optionCommentPattern` in background and build-lang.

apps/vscode/syntaxes/build-lang.js
vezwork: We now use `optionCommentPattern` which addressed both of these

apps/vscode/syntaxes/build-lang.js
vezwork: There's a way with TextMate! We now have a rule that captures option comments until there's a line that doesn't, and then all the rest of the lines get captured by another a separate regex.

apps/vscode/src/providers/background.ts
vezwork: done

apps/vscode/src/providers/background.ts
vezwork: same as other comments, done by using `optionCommentPattern`!

vezwork:
@juliasilge thanks for the great comments. I have addressed them. I have also added a setting to turn on and off the background/underline/opacity. 

I have tested various .qmd files in both VSCode and Positron. I've also tested turning the setting on and off in both. 

I think the PR is in a good spot!

apps/vscode/package.json
```diff
@@ -1067,6 +1067,13 @@
           "default": 250,
           "markdownDescription": "Millisecond delay between background color updates."
         },
+        "quarto.cells.options.background": {
+          "order": 23,
+          "scope": "window",
+          "type": "boolean",
+          "default": true,
+          "markdownDescription": "Apply a visual treatment to cell option comments (e.g. `#|` lines): a slightly darker background, dimmed text, and a separator below the options."
```
juliasilge: IIUC this setting is read inside `if (highlightingConfig.enabled())`, so `quarto.cells.background.color: off` also turns off the option background, the dimming, and the separator. Makes sense but can we clarify for folks?

```suggestion
          "markdownDescription": "Apply a visual treatment to cell option comments such as `#|` lines: a slightly darker background, dimmed text, and a separator below the options. When `#quarto.cells.background.color#` is `off`, this setting has no effect."
```

apps/vscode/package.json
```diff
@@ -1067,6 +1067,13 @@
           "default": 250,
           "markdownDescription": "Millisecond delay between background color updates."
         },
+        "quarto.cells.options.background": {
+          "order": 23,
```
juliasilge, line 1071: `order: 23` is already taken by `quarto.cells.hoverHelp.enabled` (line 998). Every other `quarto.cells.*` setting has a unique order. With the tie, this one interleaves with the hover-help group in the Settings UI instead of the background settings.

The background group runs 19 to 22, so 23 is the right slot for this setting. Can we shift the other ones:

- `hoverHelp.enabled` 23 → 24
- `signatureHelp.enabled` 24 → 25
- `diagnostics.enabled` 25 → 26
- `diagnostics.debounceDelay` 26 → 27
- `useReticulate` 27 → 28

juliasilge (approved):
This is really nice now! 🤩 

I have two comments that it would be great to incorporate before merging, but no need for me to review again unless something comes up.

PR #1094, opened 2026-08-18
positron testing: move from macos to ubuntu runners

isabelizimm:
Updates related to https://github.com/posit-dev/positron/issues/15281

Moving to more cost-effective runners, and pinning to latest stable versions of tools


juliasilge (approved):
Thank you so much! 🙌

PR #1115, opened 2026-09-07
In Positron, use virtual notebook (in memory) for LSP features instead of vdoc (on disk)

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/14540

When Positron serves embedded language features for code cells itself (the new `quarto.embeddedLanguageFeatures.native` setting, currently verified for R and Python), this extension now stands down and stops serving those features from virtual document temp files. Vanilla VS Code is untouched and every gate falls back to the vdoc path.

Per feature:

- **Completion, hover, signature help, definition** (`lsp/client.ts`): gated per language. In a cell of a natively served language the middleware returns `undefined` instead of consulting a vdoc. Two orderings are preserved deliberately: `#|` option lines still fall through to the Quarto language server (chunk-option completions come from there), and image hover in prose runs before the gate.
- **Semantic tokens** (`providers/semantic-tokens.ts`): gated per language, returning `undefined` and NOT `next()` in this case. The Quarto server's semantic-tokens capability is kind of a fake thing that answers `{ data: [] }`, and an empty stream counts as an answer that can suppress Positron's provider in the shared score group.
- **Document symbols** (`lsp/client.ts`, new `lsp/cell-symbols.ts`): on the native path the middleware fetches `positron.executeQuartoCellSymbolProvider` once per request and nests each cell's symbols under its chunk symbol by range containment. The 500 ms retry loop does not run on this path; when a language server registers, the editor re-requests symbols on its own. Headings always come from the server and `next()` is never gated here.
- **Formatting** (`providers/format.ts`): document and range formatting call `positron.executeQuartoCellFormattingProvider` / `positron.executeQuartoCellRangeFormattingProvider`. A veto (`vetoedCells > 0`) abandons the format with the existing aggregate message, matching the vdoc path's all-or-nothing semantics. `quarto.formatCell` stays vdoc-backed (this will be a follow-up).
- **Diagnostics** (`providers/diagnostics.ts`): no per-language vdoc sessions are created while native is on; toggling the setting disposes and re-arms the manager live.
- **Statement range / help topic** (`lsp/client.ts`): the Positron registrations follow the setting live via an `onDidChangeConfiguration` listener. It's disposed when native turns on, re-registered when it turns off. Last registration wins these single-answer features, so both toggle directions land correctly.

Also I'll highlight that vdoc temp file creation and deletion are now logged at debug level to the Quarto output channel (`[vdoc] Created/Deleted <path>`). This makes it a whole lot easier to do the validation below, as the vdoc files are transient dotfiles that live for milliseconds on a fast computer.

## How the gate works

`host/native-features.ts` has three parts:

1. **Capability detection is connected to command presence.** At activation the extension probes for the internal `_executeQuartoCell*` command ids with `getCommands(false)`. It must be the internal ids, because the public `positron.executeQuartoCell*` commands are API commands, which are never mirrored into the registry `getCommands` reads, and `getCommands(true)` filters underscore-prefixed ids. Positron registers these commands unconditionally, so presence exactly tracks "this build can serve this command natively"; vanilla VS Code and older Positron builds have no such commands.
2. **The setting is read live** on every call, so toggling takes effect without a window reload for every gate.
3. **The language set is extension-side** (`r`, `python`), matched against `EmbeddedLanguage.ids` so aliases count. A language not in the set keeps its vdoc. This does mean we are hard-coding which languages can use these virtual notebooks but it's the safe direction.

Both skew directions are meant to be conservative and fail safe, so old extension + new Positron and new extension + old Positron both stay on vdocs.

## Behavior changes

- **Format Document now formats every language's cells**, not just the cursor's, and every Python cell in the document, where the vdoc path deliberately formatted only the cell under the cursor. Format-on-save does the same. This is the change most likely to surprise a Python user, I think, but probably in a good way.
- **`quarto.cells.hoverHelp.enabled` / `signatureHelp.enabled` can no longer suppress anything once native is on.** Those settings gate the extension's middleware, which is no longer in the loop; Positron serves hover and signature help regardless. If we get comlaints about this, we'll need to figure out how to respect this setting in Positron.

## Languages outside the verified set

Three gates are whole-document (symbols, formatting, statement range / help topic), so a language we don't yet support, e.g. the `{sql}` chunk a knitr document can carry, is handed to Positron too. I validated against a setting-off control window, where the `{sql}` chunk's outline symbols match the control, Format Document leaves the sql chunk unchanged without blocking the R and Python cells, and Cmd+Enter/F1 behave the same in both windows. IIUC the residual risk is selector scheme where a provider registered for `{ language, scheme: 'file' }` answers a vdoc temp file but not a `vscode-notebook-cell:` URI.

## Known gap: unterminated fences

Positron's parser ignores an unclosed fence; the extension's engine parses one to the end of the document. So while a cell is being typed, or after a closing fence is deleted, completion, hover, signature help, and definition go quiet inside that cell until the fence is closed. The extension has gated off and Positron has no cell to serve. I think we should accept this for now rather than fix extension-side; the only extension-side fix is mirroring Positron's fence-closure rule, which recreates the duplication this PR exists to remove. In validation this felt pretty transient and OK in terms of user experience. Closing the fence brings features straight back. The real fix is Positron-side in the next PR over there. Either the parser can own an unterminated final cell, or Positron can expose a cell-ownership query the gate can consult.

## Release coordination

We'll need to do an extension release before the next set of changes in Positron, as the extension bump and the default flip will stick together there.

## Validating this PR

In a Positron build that carries the virtual notebook support (i.e. the current release or a daily), run this branch's extension dev host (`Run Extension` from this repo). In the dev host window, set `"quarto.embeddedLanguageFeatures.native": true`, open a `.qmd` that has R and Python chunks, a `#|` option line, and an image link in prose, and start R and Python sessions. Open the Quarto output channel and set its level to Debug. This PR adds `[vdoc] Created/Deleted <path>` logging there, which is how you can see whether the extension is serving from temp files at all.

With the setting ON:

1. The Quarto channel shows `[NativeFeatures] Host serves Quarto cell language features...` at startup. If that line is absent, the host has no cell commands and everything below is silently on the old path.
2. Hover, completion, signature help, and go-to-definition work inside R and Python chunks, and no `[vdoc]` lines appear in the log. Before this PR, each of these created a temp file.
3. Typing `#| ` in a chunk still offers chunk-option completions (those come from the Quarto language server, not the cells).
4. Hovering an `` link in prose still shows the image preview.
5. The Outline shows headings with chunks nested under them and code symbols under the chunks. (A flat "Quarto Code Cells" group that we _don't_ want also appears; removing it is Positron-side, in the next PR.)
6. Format Document formats every R and Python cell; `#|` lines, fences, and prose are untouched. Note the deliberate scope change described above. Before this PR, only the cursor's language (and for Python, only the cursor's cell) was formatted.
7. <kbd>Cmd+Enter</kbd> and <kbd>F1</kbd> work in a chunk on first window load.
8. Toggle the setting off and on WITHOUT reloading the window, and <kbd>Cmd+Enter</kbd> keeps working in both states (the statement-range and help-topic registrations follow the setting live).
9. Introduce an error in an R chunk and in a Python chunk, and there is one Problems entry each on the `.qmd`, with no duplicates.

Controls:

10. Set the setting to false and reload. Now the same interactions produce `[vdoc]` lines in the log, and everything still works. That is the old path, unchanged.
11. Run the dev host in vanilla VS Code. There is no `[NativeFeatures]` line, everything is served from vdocs, behavior is unchanged from before this PR.

### DIFF
--- a/apps/vscode/src/host/cell-features.ts
+++ b/apps/vscode/src/host/cell-features.ts
@@ -4,6 +4,39 @@
  * Copyright (C) 2026 by Posit Software, PBC
  */
 
+/**
+ * Who answers language feature requests inside Quarto code cells: this
+ * extension, from a `.vdoc.*` temp file, or the host, from an in memory
+ * virtual notebook. Both answers reach the editor for one request, so only one
+ * can answer it.
+ *
+ * This module is the entry point for that decision, and only the decision. The
+ * features stay in their own providers. Each one asks here, then stands down or
+ * carries on.
+ *
+ * Gated per language, so a cell the host does not cover keeps its virtual
+ * document:
+ *
+ * - completion, hover, signature help, go to definition (`lsp/client.ts`)
+ * - diagnostics (`providers/diagnostics.ts`)
+ *
+ * Gated per document, because one request covers the whole file and only one
+ * answer survives. Any cell the host owns hands it every cell, including ones
+ * in languages the host does not cover:
+ *
+ * - document symbols (`lsp/client.ts`)
+ * - document and range formatting (`providers/format.ts`)
+ * - semantic tokens (`providers/semantic-tokens.ts`)
+ * - statement range and help topic (`lsp/client.ts`), which are registrations,
+ *   so they are disposed and registered again rather than returning early
+ *
+ * Not gated: `quarto.formatCell` still uses a virtual document.
+ *
+ * The host owns a feature when it has the cell commands (probed at activation
+ * by {@link detectCellFeatureOwnership}), the setting is on, and, for a
+ * per-language gate, the language is one of `kHostOwnedLanguages`.
+ */
+
 import { commands, LogOutputChannel, workspace } from "vscode";
 import { tryAcquirePositronApi } from "@posit-dev/positron";
 
--- a/apps/vscode/src/host/cell-features.ts
+++ b/apps/vscode/src/host/cell-features.ts
@@ -1,5 +1,5 @@
 /*
- * native-features.ts
+ * cell-features.ts
  *
  * Copyright (C) 2026 by Posit Software, PBC
  */
@@ -14,11 +14,11 @@ import { EmbeddedLanguage } from "../vdoc/languages";
  * Positron core, not by this extension, so it is read through the full
  * configuration rather than the `quarto` section we contribute.
  */
-export const kNativeFeaturesSetting = "quarto.embeddedLanguageFeatures.native";
+export const kHostCellFeaturesSetting = "quarto.embeddedLanguageFeatures.native";
 
 /**
  * Commands whose presence says this host carries the virtual notebook (see
- * {@link detectNativeEmbeddedFeatures}).
+ * {@link detectCellFeatureOwnership}).
  *
  * These are the INTERNAL ids, and the code calls the public
  * `positron.executeQuartoCell*` ones. The internal ids are what a probe can
@@ -29,23 +29,23 @@ export const kNativeFeaturesSetting = "quarto.embeddedLanguageFeatures.native";
  * The internal commands are registered in the workbench, so they are visible,
  * as long as the probe does not filter underscore-prefixed ids.
  */
-const kNativeFeatureCommands = [
+const kCellOwnershipCommands = [
   "_executeQuartoCellSymbolProvider",
   "_executeQuartoCellFormattingProvider",
   "_executeQuartoCellRangeFormattingProvider",
 ];
 
 /**
- * Languages Positron is verified to serve natively. Matched against
+ * Languages whose cells the host is verified to own. Matched against
  * {@link EmbeddedLanguage.ids}, so an alias of a listed language counts too.
  *
  * Add one language at a time, once its cell providers have been verified end to
  * end: a document that is not covered here keeps its virtual document, which is
  * the safe direction.
  */
-const kNativeLanguages = new Set(["r", "python"]);
+const kHostOwnedLanguages = new Set(["r", "python"]);
 
-let nativeAvailable = false;
+let cellCommandsAvailable = false;
 
 /**
  * Whether this host carries the virtual notebook.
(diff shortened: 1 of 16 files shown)


juliasilge:
I started running all the Positron E2E tests with these changes, and there is a little bit of cleanup still to do on the Positron side, such as https://github.com/posit-dev/positron/issues/15974.

apps/vscode/src/host/cell-features.ts
```diff
@@ -0,0 +1,129 @@
+/*
+ * native-features.ts
+ *
+ * Copyright (C) 2026 by Posit Software, PBC
+ */
```
vezwork, line 5: It would be nice if you add a brief overview comment of what "native features" are. Is this the entry point for them?

apps/vscode/src/host/native-features.ts
```diff
@@ -0,0 +1,129 @@
...
+  if (!tryAcquirePositronApi()) {
+    nativeAvailable = false;
+    return;
+  }
+
+  // `false` keeps the underscore-prefixed ids we are looking for
+  const all = await commands.getCommands(false);
+  nativeAvailable = kNativeFeatureCommands.every((command) =>
+    all.includes(command)
+  );
```
vezwork: consider pulling out a function like

```ts
const isNativeAvailable = async () => {
  if (!tryAcquirePositronApi()) return false;

  const all = await commands.getCommands(false);
  return kNativeFeatureCommands.every((command) =>
    all.includes(command)
  );
}

File: apps/vscode/src/host/cell-features.ts [hunk 3]
```diff
@@ -0,0 +1,129 @@
...
+      `while ${kNativeFeaturesSetting} is on.`
+    );
+  } else if (
+    workspace.getConfiguration().get<boolean>(kNativeFeaturesSetting) === true
+  ) {
+    outputChannel?.warn(
+      `[NativeFeatures] ${kNativeFeaturesSetting} is on, but this host has no ` +
+      "Quarto cell commands. Serving embedded language features from virtual " +
+      "documents, which can duplicate what the host provides."
+    );
```
vezwork, line 127: nice logs. These were very helpful in testing.

apps/vscode/src/host/native-features.ts
```diff
@@ -0,0 +1,129 @@
+/*
+ * native-features.ts
```
vezwork: I would prefer a more directly descriptive term than "native-features", like "positron-supported-features" or "positron-provided-features".

vezwork: This goes for everywhere where the term "native" is used.

apps/vscode/src/lsp/client.ts
```diff
@@ -328,6 +358,11 @@ function embeddedCodeCompletionProvider(engine: MarkdownEngine) {
     const vdoc = await virtualDoc(document, position, engine);
 
     if (vdoc && !isWithinYamlComment(document, position)) {
+      // stand down when the host serves this language's cells itself
```
vezwork:
```suggestion
      // when the host is Positron, it may serve the language's cells itself and the extension should stand down (not try to provide them)
```

vezwork (commented):
Code looks good with comments and suggestions.

Tested (as in the PR description) in Positron with the setting on and off, and in VScode. VSCode behaviour seems unaltered.

In testing I noticed two things:
- image hover preview does not ever seem to work in Positron nor VSCode
- diagnostics do not show up when `native` setting is off in Positron

vezwork:
> **Document symbols** (lsp/client.ts, new lsp/cell-symbols.ts): on the native path the middleware fetches positron.executeQuartoCellSymbolProvider once per request and nests each cell's symbols under its chunk symbol by range containment. The 500 ms retry loop does not run on this path; when a language server registers, the editor re-requests symbols on its own. Headings always come from the server and next() is never gated here.

There is a 500 ms retry loop for getting document symbols? It runs on the other path? The non-native one? Does it still run for getting the headers?

apps/vscode/src/host/native-features.ts
juliasilge: Great point!

apps/vscode/src/host/cell-features.ts
juliasilge: Added!

juliasilge:
> There is a 500 ms retry loop for getting document symbols? It runs on the other path? The non-native one? Does it still run for getting the headers?

Good questions, and my wording in the description was not quite right. It is not a loop per se but it retries one time. All of this is in `embeddedDocumentSymbolProvider` (`lsp/client.ts`) and it came in with #972. `enhanceSymbolsWithCodeCellContent` walks the symbol tree, and for each chunk symbol it builds a virtual document and asks the language server for that cell's language. If any one of those servers answers `undefined`, the function returns the sentinel `'HadUndefined'`. The caller then waits 500 ms and calls it one more time. If the second attempt also reports `'HadUndefined'`, it gives up and returns `baseSymbols`.

It _only_ runs on the other vdoc path, not on the new Positron-owned path. We don't need it because we're hooked up to the `.qmd` itself and not another file on disk. It never runs on headers either.

Additionally:

- I logged https://github.com/quarto-dev/quarto/issues/1122 for us to solve the problem with the image hover.

- We have https://github.com/posit-dev/positron/issues/15654 and https://github.com/posit-dev/positron/issues/16015 to fix up some remaining problems with diagnostics. I _believe_ 🤞 we have things set up correctly here in the extension but I won't absolutely promise I won't be back to fix up something else here. 😅

vezwork (approved):

PR #1116, opened 2026-09-08
Respect `quarto.cells.useReticulate` setting in Positron

juliasilge:
Addresses https://github.com/posit-dev/positron/issues/6675 in part

This PR fixes a bug in Positron. Before this change, Positron ignored the `quarto.cells.useReticulate` setting. When you ran a Python cell in a document with `engine: knitr`, Positron always rewrote the code to run in the R console through `reticulate::repl_python(...)`. The setting had no effect. Now the Positron cell execution path checks the same setting that the default VS Code host already uses.

I also added a regression test. The test runs a knitr Python cell with the setting off. It shows that the cell executes in the Python console instead of R.

## Known limitation

This change does not solve the full problem in the issue about the difference between running a statement and Run Cell. In Positron, <kbd>Cmd+Enter</kbd> (run the current statement) sends the command to Positron core's `positron.executeCodeFromPosition` command. This command always runs Python cells on the Python console. It does not check the `quarto.cells.useReticulate` setting, and this extension has no hook to change that path. As a result, with the setting on, Run Cell now goes to R through reticulate, but Run Line still goes to Python. A fix for that part of the issue needs a change in Positron itself.

## Validation steps

- `yarn test-positron` passes (includes the new test)

- With a `.qmd` file that has both an `r` chunk and a `python` chunk:

  - With the setting on (the default), running a knitr Python cell with <kbd>Cmd+Shift+Enter</kbd> runs it via R through reticulate:

Screenshot 2026-09-08 at 3 55 54 PM

  - With the setting off, the same cell runs directly in the **Python** console:
 
Screenshot 2026-09-08 at 4 04 24 PM

  - R cells run the same way, with the setting on or off

  - Switching the setting on and off again repeats this same behavior each time

### DIFF
--- a/apps/vscode/CHANGELOG.md
+++ b/apps/vscode/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## 1.138.0 (Unreleased)
 
+- In Positron, running a Python cell in a knitr document now respects the `quarto.cells.useReticulate` setting, instead of always routing it through reticulate on the R console (<https://github.com/quarto-dev/quarto/pull/1116>).
 
 ## 1.137.0 (Release on 2026-09-04)
 
--- a/apps/vscode/src/host/positron.ts
+++ b/apps/vscode/src/host/positron.ts
@@ -95,8 +95,13 @@ export function positronExtensionHost(outputChannel?: vscode.LogOutputChannel): 
                 return;
               }
 
-              if (language === "python" && isKnitrDocument(document, engine)) {
-                language = "r";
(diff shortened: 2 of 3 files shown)


vezwork (approved):
The code change is simple and good. Tested in Positron daily confirmed that by default (when the useReticulate setting is on) python cells in a qmd with an r cell get sent to the r console with reticulate. When the setting is off, they get sent to the python console.

Looking forward to command+enter being fixed to respect the setting as well.

Issue #1119, opened 2026-09-09
Local Zotero library: new or edited items do not appear until Zotero checkpoints its WAL file


juliasilge:
### Bug description

The Quarto extension reads the local Zotero library. It copies `zotero.sqlite` and queries the copy. Zotero writes new and changed items to a `-wal` file first. Zotero moves this data into the main `zotero.sqlite` file only at intervals. This process is a checkpoint.

Quarto's staleness check compares only the mtime of the main `zotero.sqlite` file. As a result, it does not detect newer data in the `-wal` file.

Because of this, an item that you add or edit in the Zotero desktop app does not appear in the Visual Editor's Insert Citation picker. It appears only after Zotero checkpoints the WAL on its own. Zotero checkpoints on idle time, at a size limit, or when you quit Zotero.

## Steps to reproduce

1. Set `Quarto > Zotero > Library` to `local`.
2. Make sure that the citation picker works. See #1101.
3. Add a new item to your Zotero library in the Zotero desktop app.
4. In the Visual Editor, open Insert > Citation.
5. Search for the new item.

Result: the new item does not appear. Zotero shows the new item at once in its own library view.

## Evidence

```
$ stat -f "%N  mtime=%Sm" ~/Zotero/zotero.sqlite ~/Zotero/zotero.sqlite-wal
/Users/juliasilge/Zotero/zotero.sqlite      mtime=Sep  9 11:58:55 2026
/Users/juliasilge/Zotero/zotero.sqlite-wal  mtime=Sep  9 12:33:24 2026
```

The `-wal` file is 35 minutes newer than the main file. At that time, the new item still did not appear. This shows that the new data was in the WAL file, unread.

## Root cause

`node-sqlite3-wasm` is the library that `packages/editor-server/src/core/zotero/local/db.ts` uses to read the local database. This library cannot open a WAL-mode database. Direct tests show this, even when the `-wal` and `-shm` companion files are next to the main file.

Zotero uses WAL mode by default in current versions.

To work around this, the copy-then-read path in `db.ts` changes the SQLite file-header bytes that mark WAL mode. This change lets Quarto open the copy (see the fix for #1101).

This workaround reads only the data that Zotero has already checkpointed into the main file. It cannot read data that exists only in the `-wal` file.

## Possible directions

[...]

PR #1120, opened 2026-09-09
Fix Zotero silently disappearing as a citation source

### LINKED ISSUE #1101: Zotero <-> Visual Editor link ?broken in 1.136.0
### Bug description

Hello,

Thank you for the very useful connection between Positron Quarto Editor with Zotero - I'm a global health PhD researcher, and this connection means my entire research workflow stays in Positron. 

I recently found the local Zotero (10) connection in 1.136.0 wasn't working. Zotero doesn't show in the list of sources. I have found a work around. In summary - Zotero 10 appears not to connect with any recent version of Quarto. Zotero 9 appears not to connect with the most recent version of Quarto.

My observations:
1. Web connection appears to work (although variably, plus I work in a place with internet constraints)
2. Rolling back Zotero 10 to Zotero 9 did not resolve the issue using version 1.136.0
3. Rolling back Quarto Extension to 1.135.0 whilst using Zotero 9 restored the local connection (hooray!)
4. Updating to Quarto 1.136.0 whilst using Zotero 9 replicated the issue
5. Returning to Zotero 10 replicated the issue on any Quarto version over the past year

Please let me know if any other info is useful.

### Steps to reproduce

1. Open Zotero 9 or Zotero 10
2. Open Positron with Quarto Extension version 1.136
3. Check Positron Settings > Quarto > Zotero is set to local
4. Check Zotero > Preferences > Advanced > "allow other applications" is checked
5. Save a test.qmd file
6. Right click to enter visual mode
7. Insert > Citation

### Actual behaviour

Zotero does not appear as a source

### Expected behaviour

Zotero local library does appear a source

### Your environment

Positron 2026.08.2 build 4 on macOS Tahoe

juliasilge:
Fixes #1101

Zotero could disappear as a citation source in the visual editor, without any error message. This affected the local library, and in one case, newly added items. Four separate bugs caused this, all of which are addressed here:

1. **Config push raced LSP startup.** The initial Zotero configuration push to the language server could race the server's own startup. This was a regression in 1.136.0, when LSP startup became lazy. A failed push now retries. The Quarto output channel logs each failure. If the sync still fails, Quarto shows a warning message with a retry option.
2. **Config changes did not retry on failure.** A change to `Quarto > Zotero > Library`, the data directory, or the API key pushed the new config one time only. On failure, the extension discarded the result. The LSP then kept the old settings for the rest of the session, with no warning. This now uses the same retry path as the initial sync.
3. **Concurrent lookups raced on the same cached database copy.** The citation picker sends `getCollections` and `getActiveCollectionSpecs` at the same time. Both requests copy the local Zotero database to the same cache path. A failed open in one request could delete the copy that the other request still needed. As a result, both requests failed with `unable to open database file`. Access to the cached copy is now one request at a time, per data directory.
4. **The local database reader could not open a WAL-mode library.** Current Zotero versions (9 and 10) use WAL mode by default. `node-sqlite3-wasm` cannot open a WAL-mode database. We now clear the WAL flag in the file header of the throwaway copy before we open it.

## Changes I decided NOT to make here

These are remaining problems, but I think too risky to try to address now to fix the 1.136.0 regression:

- **Recreate the LSP client after a startup failure.** An automated review found a problem inside the `vscode-languageclient` library itself. Its `LanguageClient.start()` method saves its internal start promise and never clears it after a `StartFailed` state. Our retry logic cannot recover from this specific failure for that reason. A correct fix needs a new `LanguageClient` instance. That change touches shared LSP startup code for every LSP feature, not only Zotero. Let's leave this alone for now as it is a bigger and more risky change than the rest of this PR.
- **Register LSP custom methods earlier during server startup.** The client reports the state "Running" as soon as the LSP handshake completes, but the server does not register its custom RPC methods, including the Zotero config push, until later in its own `onInitialized` handler. An earlier registration can shrink this race, or maybe remove it, at the source. This also touches shared LSP startup code so let's leave it for now. The retry from fix 1 already covers the common case.
- **Read Zotero's `-wal` journal file.** Zotero writes new and changed items to a `-wal` file first. Zotero moves this data into the main `zotero.sqlite` file only at a checkpoint. Our fix makes the main file open again, but an item added or edited in the Zotero desktop app still does not appear in the citation picker until Zotero checkpoints on its own. A correct fix needs either the `sqlite3` command-line tool to force a checkpoint, or a different SQLite library that supports WAL mode natively. Both options are bigger changes than this PR should take on; see the separate issue #1119.

## Manual validation

I tested this against a real local Zotero 10 library:

Screenshot 2026-09-09 at 12 49 52 PM

1. Insert Citation shows the local library as a source. The inserted citation shows the correct author, year, and title.
2. Open the citation picker many times and insert several citations in a row. No `unable to open database file` error appears in the Quarto output channel.
3. Change `Quarto > Zotero > Library` between `none` and `local` while the extension was running. Each change syncs the config correctly.
4. A full reload of the window, a cold start, still activates Zotero correctly.
5. Known limitation, not fixed in this PR, is that an item added or edited in the Zotero desktop app does not appear in the citation picker until Zotero checkpoints its WAL file on its own.


jmcphers (approved):
Image

Working great!

PR #1121, opened 2026-09-09
Enable Render on Save toggle in visual editor

juliasilge:
The **Render on Save** checkbox appears in the visual editor's action bar, but it wasn't correctly working there. This PR makes the checkbox (more) correctly functional in the visual editor. I'll be honest that there is still some weirdness with the `quarto.render.renderOnSave` setting, but this is an improvement.

This is the Quarto-side follow-up called out in https://github.com/posit-dev/positron/pull/16022, which is making extension-contributed action bar checkboxes follow their command's `enablement` expression. cc @dhruvisompura

## Root cause

Two issues:

1. The `quarto.toggleRenderOnSave` command's `enablement` was `editorLangId == quarto`, which doesn't match when the visual editor (a custom editor) is active so the checkbox was not fully/correctly enabled there.
2. The `quarto.editor.type` / `quarto.editor.renderOnSave` / `quarto.editor.renderOnSaveShiny` context keys that drive the checkbox's checked state were only updated via `onDidChangeActiveTextEditor`, which custom editors don't fire, so the state was stale even where the checkbox was visible.

## Changes

- `package.json`: add `activeCustomEditorId == 'quarto.visualEditor'` to the command's `enablement`, matching the pattern used by other menu contributions.
- `editor.ts`: add an `onDidChangeActiveEditor` event to the visual editor tracker, fired from each tracked webview panel's `onDidChangeViewState` when its panel becomes active, and expose it via `VisualEditorProvider.onDidChangeActiveEditor()`.
- `context-keys.ts`: refactor `setEditorContextKeys` / `setLanguageContextKeys` to take a `TextDocument` (they only ever used the document), subscribe to the new event so context keys refresh when a visual editor gains focus, and fall back to the active visual editor's document in the configuration-change listener.

## Validation

- Open a `.qmd` in visual mode: the checkbox appears and is clickable.
- Toggle the checkbox in visual mode and save: render kicks off when checked, doesn't when unchecked.
- Toggle in visual mode, switch to source mode (and back): state carries across.
- Regression: source-mode behavior unchanged; checkbox stays hidden for non-Quarto files.

### DIFF
--- a/apps/vscode/src/providers/editor/editor.ts
+++ b/apps/vscode/src/providers/editor/editor.ts
@@ -26,7 +26,9 @@ import {
   Selection,
   TextEditorRevealType,
   GlobPattern,
-  TabInputText
+  TabInputText,
+  Event,
+  EventEmitter
 } from "vscode";
 
 import { projectDirForDocument, QuartoContext } from "quarto-core";
@@ -266,6 +268,11 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
 
   public static recordPendingSwitchToVisual(document: TextDocument) {
     this.editorPendingSwitchToVisual.add(document.uri.toString());
+  }
+
+  // fires when a visual editor's webview panel becomes active
+  public static onDidChangeActiveEditor(): Event<TrackedEditor> {
+    return this.visualEditors.onDidChangeActiveEditor;
   }
 
   public static activeEditor(includeVisible?: boolean): QuartoVisualEditor | undefined {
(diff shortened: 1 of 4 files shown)


apps/vscode/src/providers/editor/editor.ts
```diff
@@ -268,6 +270,11 @@ export class VisualEditorProvider implements CustomTextEditorProvider {
     this.editorPendingSwitchToVisual.add(document.uri.toString());
   }
 
+  // fires when a visual editor's webview panel becomes active
+  public static onDidChangeActiveEditor(): Event<TrackedEditor> {
+    return this.visualEditors.onDidChangeActiveEditor;
+  }
```
vezwork:
```suggestion
private static onDidChangeActiveEditor = VisualEditorProvider.visualEditors.onDidChangeActiveEditor
```

nitpick. The PR's code here makes for a call-site `VisualEditorProvider.onDidChangeActiveEditor()(editor => { .. })` which structurally concerned me because every invocation of `VisualEditorProvider.onDidChangeActiveEditor()` could be returning a new function (which it isn't, and even if it was it wouldn't be a problem since I think its only ever called once, but.... its a mild code smell to me )

vezwork (approved):
Validated in Positron and approved with a nitpick

PR #1123, opened 2026-09-10
Fix LSP crash on unparseable cell option YAML by handling null lint result

### LINKED ISSUE #1118: Quarto-LSP crash due to bad indentation
### Bug description

Here is a repro for a Quarto-LSP crash reported in https://github.com/posit-dev/positron/issues/15965#issuecomment-5588268127:

````md
---
title: "Quarto YAML lint crash repro"
---

Reproduces the Quarto LSP crash from
https://github.com/posit-dev/positron/issues/15965#issuecomment-5588268127:

    Error found during linting [YAMLException: bad indentation of a mapping entry (undefined, 205:157)
    TypeError: Cannot read properties of null (reading 'map')
        at y0e (.../lsp.js:367:21130)
        at async Object.computeOnSaveDiagnostics (.../lsp.js:373:31225)

Steps:
1. Open this file in Positron with the Quarto extension active.
2. Save it (Cmd+S). The crash happens in `computeOnSaveDiagnostics`, so
   opening/editing alone may not trigger it.
3. Check Output -> Quarto for the YAMLException + TypeError above. After a
   few crashes in quick succession you should also see:
     [Error] The Quarto LSP server crashed 5 times in the last 3 minutes.
     The server will not be restarted.

```{mermaid}
%%| fig-align: center
%%| fig-cap: Sentiment analysis workflow used in the study: results and discussion of findings across multiple datasets and models tested here.
flowchart LR
  A[Data] --> B[Model]
  B --> C[Results]
```

````

Looks like a missing `null` check in https://github.com/quarto-dev/quarto/blob/b81cfe74ca6d7249b807621b8c9da27f7cbdeab8/apps/lsp/src/service/providers/diagnostics-yaml.ts#L29-L30

### Steps to reproduce

_No response_

### Actual behavior

_No response_

### Expected behavior

_No response_

### Your environment

_No response_

### PR DESCRIPTION
**Pull request by juliasilge**
Fixes #1118 (the upstream report is from posit-dev/positron#15965)

This PR fixes a Quarto LSP server crash that occurred when saving a document containing unparseable YAML in code cell options (for example, a `%%| fig-cap:` value containing an unquoted colon). After five crashes in three minutes, the client stops restarting the server and all language features are lost until reload.

## Root cause

When cell option YAML fails to parse, `getLint()` in the bundled `vs-code.mjs` module catches the resulting `YAMLException`, logs "Error found during linting", and resolves to `null`. The on-save diagnostics path in `provideYamlDiagnostics` then called `.map()` on that `null` value, throwing a `TypeError` that crashed the language server:

```
Error found during linting [YAMLException: bad indentation of a mapping entry ...]
TypeError: Cannot read properties of null (reading 'map')
    at computeOnSaveDiagnostics (.../lsp.js)
```

The other caller of `getYamlDiagnostics` (in `apps/lsp/src/custom.ts`) already handled `null` correctly, which is why this only crashed on save.

I verified against the reproduction document in #1118, and now saving no longer crashes the server.

## A possible follow-up

With this fix, a document with broken cell option YAML silently produces no YAML diagnostics on save (the parse error is only logged to the Output channel). A future improvement could surface the parse error as a diagnostic using the position info in the `YAMLException`; that requires changes in the Quarto CLI, where the exception is currently swallowed. The null guard here is still valuable regardless.

### DIFF
--- a/apps/lsp/src/quarto.ts
+++ b/apps/lsp/src/quarto.ts
@@ -38,7 +38,7 @@ export interface Quarto extends QuartoContext {
     token: AttrToken,
     context: EditorContext
   ): Promise<CompletionItem[]>;
-  getYamlDiagnostics(context: EditorContext): Promise<LintItem[] | null>;
+  getYamlDiagnostics(context: EditorContext): Promise<LintItem[]>;
   getHover?: (context: EditorContext) => Promise<HoverResult | null>;
 }
 
@@ -50,7 +50,9 @@ export async function initializeQuarto(context: QuartoContext): Promise<Quarto> 
     getAttrCompletions: initializeAttrCompletionProvider(
       context.resourcePath
     ),
-    getYamlDiagnostics: quartoModule.getLint,
+    // the external getLint() resolves to null when linting fails
+    getYamlDiagnostics: async (context) =>
+      (await quartoModule.getLint(context)) ?? [],
     getHover: quartoModule.getHover
   };
 
--- a/apps/lsp/src/quarto.ts
+++ b/apps/lsp/src/quarto.ts
@@ -38,7 +38,7 @@ export interface Quarto extends QuartoContext {
     token: AttrToken,
     context: EditorContext
   ): Promise<CompletionItem[]>;
-  getYamlDiagnostics(context: EditorContext): Promise<LintItem[]>;
+  getYamlDiagnostics(context: EditorContext): Promise<LintItem[] | null>;
   getHover?: (context: EditorContext) => Promise<HoverResult | null>;
 }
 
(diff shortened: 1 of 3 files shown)


lionel- (commented):
Since having to handle nulls everywhere is a bit of a footgun, or at least impractical, I'd normalise at the boundary by wrapping `getLint()` with ` ?? []` here:

https://github.com/quarto-dev/quarto/blob/73a528de8ae7f92e6108b50b7e19e6362eb411a6/apps/lsp/src/quarto.ts#L53

Then we'd keep the current non-nullable signature in the `Quarto` interface.

WDYT?

juliasilge:
Oh, that's a better idea, @lionel-. Reworked in de07dbfcf8307272c0eeded6fea30549b73819d0:

- Reverted the call-site guard in `diagnostics-yaml.ts` and the nullable signature on the public `Quarto` interface
- `initializeQuarto` now wraps the external `getLint()` with `?? []` to normalize at the boundary
- Kept `Promise<Array<LintItem> | null>` on the private `QuartoYamlModule` interface

lionel- (approved):
LG!

PR #1124, opened 2026-09-11
prepare changelog for release

vezwork:


vezwork:
Just realized I used a branch name that is not correct. It should have been release-2026-09-11 for the record. Mistake! Not consequential, just thought I should note to avoid any possible confusion.

PR #1129, opened 2026-09-17
ci: pin GitHub Actions to commit SHAs, fix Node 20 deprecation

cderv:
actions/checkout@v4 and actions/setup-node@v4 still declare `node20` in their action metadata, so GitHub Actions runners now force them onto Node 24 with a deprecation warning (see [changelog](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)). actions/upload-artifact@v4 has the same issue.

Bumps checkout, setup-node, and upload-artifact to their latest Node 24-native majors, and pins every action reference across the four workflows to a commit SHA (with a version comment) instead of a mutable tag or branch.


juliasilge (commented):
I tend to have a pretty good experience just using tags like `v7` in GH actions for projects that aren't super duper big or complicated. We don't have dependabot set up here at all for this, so if we merge this PR as is we'll go stale really fast with fixes and such for these actions. What do you think about just using the tags?

Alternatively we could also add dependabot config to this repo and set it up to be able to stay current.

cderv:
> What do you think about just using the tags?

I am ok with tags too. I thought this was becoming good practice to use the hash - and easier than before when you have 🤖 to help you update. 

I'll put back the tags and we can change in another PR for dependabot if we don't want to end in this situation of using v4 while v7 is out.

juliasilge (approved):
Thank you so much!

